PR #45, opened 2012-02-15
Catch raw hrefs as bad, warn unless wrapped in a "safe" method

oreoshake:
Add concept of a safe-ening method to mark hrefs as safe

Feature:
Warn when using unsafe hrefs.  This is a very specific case that as of now produces a ton of noise.  This came out of an xss vuln where the value was escaped but still vulnerable.

```
link_to 'asdf', h(@scary)
```

where

```
@scary = 'javascript:alert(1)'
```

or

```
@scary = 'data:  # http://palpapers.plynt.com/issues/2010Oct/bypass-xss-filters/
```

This branch accomplishes slightly intelligent warnings by adding a new command line option to declare methods that make a string URL safe (unless there is already a standard one out there).  e.g.:

```
$ brakeman . --url-safe-methods ensure_valid_protocol!

link_to 'asdf', ensure_valid_protocol!(@scary, :javascript)
```

### DIFF
--- a/lib/brakeman/checks/check_link_to_href.rb
+++ b/lib/brakeman/checks/check_link_to_href.rb
@@ -35,6 +35,7 @@ class Brakeman::CheckLinkToHref < Brakeman::CheckLinkTo
     call = result[:call] = result[:call].dup
     @matched = false
     url_arg = process call[3][2]
+    return if sexp?(url_arg) && url_arg.node_type == :string_interp && !url_arg[1].chomp.empty?
     type, match = has_immediate_user_input? url_arg
 
     if type


tomekr:
This check is flagging the following code as a finding:

```ruby
link_to("Show", current_user.bars.find(params[:id]))
```

However, this wouldn't allow a user to specify their own protocol and so isn't an actual vulnerability. As a result, I can't really use `render_safe_methods` in this scenario but I wouldn't want to disable this check completely. What would the recommended course of action be here?

presidentbeef:
@tomekr #742 should address that issue.

tomekr:
@presidentbeef Yep, that did it. Thanks!

PR #397, opened 2013-09-30
Add Support For Rails Engines

phene:
This adds support for scanning engine paths for controllers, models, views, etc.  Includes new `rails4_with_engines` test to back it up.


ilyakatz:
What do you think about making this configurable? So instead of hardcoding `engines` user can specify the name of the directory where engines reside?

presidentbeef:
Yes, see #664.

PR #529, opened 2014-07-24
demonstrated false negative by using simplified rendering syntax

brad-werth:
Hi, thanks for the great tool. Please let me know if I can provide any additional information of assistance in this matter.

### DIFF
--- a/test/apps/rails3.1/app/views/users/index.html.erb
+++ b/test/apps/rails3.1/app/views/users/index.html.erb
@@ -12,7 +12,7 @@
     <th></th>
   </tr>
 
-  <%= render 'user', :collection => @users %>
+  <%= render @users %>
 </table>
 
 <br />


brad-werth:
@presidentbeef I found some more false negatives. Is it better to report a false negative via PR with tests and code, or to simply create an issue explaining the situation (perhaps with a link to the code in my own repo)? 

Also, I was wondering if you've been able to think of any way around this issue? 

Thanks!

presidentbeef:
As long as there is code to reproduce the issue, that's the main thing.

It's possible we could cover this issue in some cases.

mltsy:
Oh, you're right, my issue is totally a duplicate of this PR :) (I guess I didn't search the PRs!)  Also, funny enough, I work with this guy - hey @brad-werth :laughing:

Is it possible to just check if the class of the argument is a String or Symbol to see if it's referring to a template name or an object? (it looks like [that's what Rails does](https://github.com/rails/rails/blob/4-0-stable/actionpack/lib/abstract_controller/rendering.rb#L165))

Or is the class of the variable not easy to determine?  I tried to follow [the brakeman code that figures out what kind of argument it is](https://github.com/presidentbeef/brakeman/blob/v3.4.1/lib/brakeman/checks/check_render.rb#L18), but I got confused, because it seems to ignore the argument altogether unless it's one of `[:partial, :template, :action, :file]` - I must have misinterpreted how it determines that... I think it [determines the render_type here](https://github.com/presidentbeef/brakeman/blob/8fa2fd13745097e0a4c4a6035d68a808b52e37ee/lib/ruby_parser/bm_sexp.rb#L529), which looks like it just returns perhaps the second "node" of the "expression"?  Which in this case would be... a variable name(?).  So that's where I got lost, because none of [the cases](https://github.com/presidentbeef/brakeman/blob/v3.4.1/lib/brakeman/checks/check_render.rb#L19) seem to handle a variable name? :confused:

presidentbeef:
In this case a bare `render ...` gets converted to `render :action => ...` In fact calls to `render` get converted to `s(:render, ...)` s-expressions in [this bit of code](https://github.com/presidentbeef/brakeman/blob/8fa2fd13745097e0a4c4a6035d68a808b52e37ee/lib/brakeman/processors/base_processor.rb#L204).

However, that's not really related. What we care about is the argument, as you mention. Brakeman performs basic data flow analysis ([some slides here](https://docs.google.com/presentation/d/1tSBuM5PzHEvqu6qR6c5548tXCWHqkvsqDKD2E5g-r84/pub?start=false&loop=false&delayms=3000)) so if there is an assignment in scope it will propagate that assignment to uses of the variable. For example we might know, from your code, that `render @client` is actually `render Client.includes(:domain).where("clients.name LIKE ? OR domains.name LIKE ?",
  "%#{params[:search]}%", "%#{params[:search]}%")`.

I believe the approach I was considering is just checking for "collectiony" type values like yours. So calls on a model that may/probably result in a collection, as opposed to a single value.

mltsy:
Oooh, I see, yeah, so it's determining that the render type is an "action" on [line 236](https://github.com/presidentbeef/brakeman/blob/8fa2fd13745097e0a4c4a6035d68a808b52e37ee/lib/brakeman/processors/base_processor.rb#L236) right?  But if the argument is neither a string/symbol, nor a hash (which is the case line 236 represents right?)... then it's actually just an object, which is when Rails will treat it as an object render, so shouldn't the render type be something else, like... perhaps `:object`?  And in the case of an "object" render, you could just skip those checks that check how the value is constructed (since they are only relevant in cases where the value is used to determine the template) right?

presidentbeef:
> then it's actually just an object

Not exactly. It could really be anything, it's just that the literal value is not able to be determined just by looking at the source code.

mltsy:
Oh, so even failing `string?` and `is_a? String` can't tell for sure that it isn't a String?  Dang.  That does make it harder... ;)

Also, wait, so this PR/issue actually *was* for a false negative (an issue not being reported properly) - mine was for an issue that shouldn't be reported in this case and is.  I guess they're not the same issue after all?

presidentbeef:
They are different aspects of the same issue. If we know it's a collection, we won't warn about dynamic render paths, but also render it as a collection in views.

mltsy:
:+1:

PR #602, opened 2015-01-04
Expand slim version support to latest (v3.0.1 at present).

### LINKED ISSUE #601: Slim Requirment lower than Rails 4.2
Bundler could not find compatible versions for gem "slim":
  In Gemfile:
    brakeman (~> 3.0.0) ruby depends on
      slim (< 3.0, >= 1.3.6) ruby

    slim-rails (~> 3.0.1) ruby depends on
      slim (3.0.0)

If possible this should be updated to use the latest version of slim which is needed for good Rails 4.2 support.

cwest:
Tests passed locally. Run against a Rails 4.2 and Rails 4.1 projects; Got expected results.

I'm not claiming a perfect understanding but things seem to be in good working order. Please let me know what I missed. :smile: 

Fixes #601.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,15 @@ From source:
     gem build brakeman.gemspec
     gem install brakeman*.gem
 
+## For Slim Users
+
+[Slim v3.0.0](https://github.com/slim-template/slim/blob/master/CHANGES#L12) dropped support for Ruby 1.8.7. Install a version of [`slim`](http://slim-lang.com/) compatible with your Ruby.
+
+| Ruby Version |       `Gemfile`       |              Command Line              |
+|--------------|-----------------------|----------------------------------------|
+| Ruby 1.8.7   | `gem 'slim', '< 3.0'` | `$ gem install slim --version '< 3.0'` |
+| Ruby 1.9+    | `gem 'slim'`          | `$ gem install slim`                   |
+
 # Usage
 
     brakeman [app_path]
--- a/Gemfile
+++ b/Gemfile
@@ -3,3 +3,10 @@ source "https://rubygems.org"
 gemspec :name => "brakeman"
 
 gem "rake", "< 10.2.0"
+
+# Slim v3.0.0 dropped support for Ruby <1.9.2.
+if RUBY_VERSION < "1.9.2"
+  gem "slim", ">=1.3.6", "< 3.0"
+else
+  gem "slim", ">=1.3.6"
+end
--- a/brakeman.gemspec
+++ b/brakeman.gemspec
@@ -23,6 +23,5 @@ Gem::Specification.new do |s|
   s.add_dependency "erubis", "~>2.6"
   s.add_dependency "haml", ">=3.0", "<5.0"
   s.add_dependency "sass", "~>3.0"
-  s.add_dependency "slim", ">=1.3.6", "<3.0"
   s.add_dependency "multi_json", "~>1.2"
 end
--- a/lib/brakeman.rb
+++ b/lib/brakeman.rb
@@ -420,7 +420,7 @@ module Brakeman
       require name
     rescue LoadError => e
       $stderr.puts e.message
-      $stderr.puts "Please install the appropriate dependency."
+      $stderr.puts "Please install the appropriate dependency: #{name}."
       exit! -1
     end
   end


cwest:
Please review @ramaboo, @presidentbeef.

cwest:
I see one of the things I missed. [The removal of Ruby 1.8.7 support from slim 3.0.0.](https://github.com/slim-template/slim/blob/master/CHANGES#L12)

cwest:
Updated to consider our `RUBY_VERSION` to maintain consistent slim support.

presidentbeef:
Should probably be `RUBY_VERSION < "1.9.0" ` or something similar just for safety.

presidentbeef:
Also I'm not sure it needs to be quite so cautious on the Slim version, or else this will need to change again at the next Slim release.

cwest:
I agree. I also think the ternary operator doesn't convey the reason for the check, so I've made it a little more clear.

cwest:
I kept it cautious based on https://github.com/presidentbeef/brakeman/pull/353#issuecomment-19931105.

If you're cool with being more forgiving so am I. Update forthcoming. :smile:

presidentbeef:
Sadly, I don't think this will work because the `RUBY_VERSION` is checked at gem creation time, not at install.

cwest:
Heh. That'd be a problem indeed. :-)

bf4:
That's true. The best you could do is in to add a post_install message to the gemspec and in the lib/brakeman.rb do something like:

```ruby
# Slim 3.0.0 dropped support for Ruby 
if RUBY_VERSION < "1.9.0"
  gem "slim", ">=1.3.6", "<3.0"
else
  gem "slim", ">=1.3.6"
end
require 'slim'
```

Why is the dependency so open-ended?

cwest:
So we have an issue here and @bf4 is right. I wanted to change the versions during install time but that seems to be a thing rubygems doesn't provide. I'm going to give his suggestion a whirl.

cwest:
Ok, it works.

cwest:
We can't remove `slim` from the gemspec because `bundle exec` explicitly disallows loading gems that aren't listed, which fails tests.

We could write another gemspec for Ruby 1.8.7 and build and distribute that version of brakeman?

It's a PITA that our upstream dependency stopped supporting a Ruby version we still want to support.

Getting food…

cwest:
And we're back. I think I like this solution. The project `Gemfile` specifies slim, and can do so on a per-ruby basis, but the gemspec no longer does.

Tested this on a Rails 4.2 app (Ruby 2.2.0) using slim 3.0.1 after installing it and upgrading my `Gemfile.lock`. Worked as expected!

presidentbeef:
Essentially this is the same as removing Slim from from the gem dependencies, which is fine I suppose.

cwest:
Right, but keeps it available for testing and development (within bundler contexts).

cwest:
This is very much how tilt deals with things, as @bf4 was trying to point out, I think. :smile: 

BTW, my apologies for exposing you to some of the more exotic attempt I made. It was pretty rough there for a bit.

cwest:
Let me know if there's anything I missed on this PR? I'd be nice for Slim and Rails 4.2 users to have the latest Brakeman. :smile:

README.md
```diff
@@ -40,6 +40,14 @@ From source:
     gem build brakeman.gemspec
     gem install brakeman*.gem
 
+## For Ruby 1.8.7 Users
+
+Slim v3.0.0 dropped support for Ruby 1.8.7. If you're using `brakeman` on Ruby 1.8.7 you will have to pre-install a version of `slim` compatible with Ruby 1.8.7:
+
+    $ gem install slim --version '~> 2.0.0'
+
+If you're on a more recent Ruby this step is _not_ required.
```
presidentbeef: This is a little misleading...you need to install slim manually if you want to process slim templates, either way.

cwest:
Docs update to look like this:

## For Slim Users

[Slim v3.0.0](https://github.com/slim-template/slim/blob/master/CHANGES#L12) dropped support for Ruby 1.8.7. Install a version of [`slim`](http://slim-lang.com/) compatible with your Ruby.

| Ruby Version |       `Gemfile`       |              Command Line              |
|--------------|-----------------------|----------------------------------------|
| Ruby 1.8.7   | `gem 'slim', '< 3.0'` | `$ gem install slim --version '< 3.0'` |
| Ruby 1.9+    | `gem 'slim'`          | `$ gem install slim`                   |

presidentbeef:
Thanks for your patience!

cwest:
@presidentbeef++ :smile:

Lordnibbler:
@presidentbeef When can we expect a new release on rubygems for this?

presidentbeef:
@Lordnibbler I don't like to be too specific about release dates, but I do expect another release relatively soon.

PR #605, opened 2015-01-06
process inner body expressions to trigger interpolated string warnings

ptoomey3:
We had added library processing to our internal fork of brakeman and noticed that some things, like warnings related to interpolated strings, were not getting triggered in scanned libs. So, I made a change in https://github.com/github/brakeman/pull/10 that processes the inner body expression similar to how other processors do (controllers, etc). However, I'm still not 100% sure if the approach taken is idiomatic with the rest of brakeman. So, please let me know if there is a more "correct" way of ensuring that the nested body's are processed so that interpolated strings are analyzed correctly. 

/cc @gregose @mastahyeti @oreoshake

### DIFF
--- a/test/apps/rails4/lib/sweet_lib.rb
+++ b/test/apps/rails4/lib/sweet_lib.rb
@@ -2,4 +2,9 @@ class SweetLib
   def do_some_cool_stuff bad
     `ls #{bad}`
   end
+
+  def test_find_group
+    #Should warn, no escaping done for :group
+    User.find(:all, :conditions => "title = 'blah'", :group => "something, #{params[:group]}")
+  end
 end
--- a/test/apps/rails4/lib/sweet_lib.rb
+++ b/test/apps/rails4/lib/sweet_lib.rb
@@ -5,6 +5,6 @@ class SweetLib
 
   def test_find_group
     #Should warn, no escaping done for :group
-    User.find(:all, :conditions => "title = 'blah'", :group => "something, #{params[:group]}")
+    system("rm #{@bad}")
   end
 end
--- a/test/tests/rails4.rb
+++ b/test/tests/rails4.rb
@@ -16,7 +16,7 @@ class Rails4Tests < Test::Unit::TestCase
       :controller => 0,
       :model => 1,
       :template => 2,
-      :generic => 47
+      :generic => 48
     }
   end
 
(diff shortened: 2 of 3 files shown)


test/tests/rails4.rb
```diff
@@ -555,7 +555,7 @@ def test_no_sql_injection_due_to_skipped_filter
   def test_command_injection_in_library
     assert_warning :type => :warning,
       :warning_code => 14,
-      :fingerprint => "9a11e7271784d69c667ad82481596096781a4873297d3f7523d290f51465f9d6",
+      :fingerprint => "21857e8872d187312a0b2470876bf6c8a8885df84c510d766f4639d95ae7cef7",
```
ptoomey3: Changed because the nested body parsing transforms the code string for the warning changes from:

```
s(:evstr, s(:lvar, :bad)))[[:class, :SweetLib]
```

to

```
s(:string_eval, s(:lvar, :bad)))[[:class, :SweetLib]
```

which affects the fingerprint.

presidentbeef:
This is probably a good idea in general, but have you checked if Brakeman 3.0 fixes the interpolated strings issue?

ptoomey3:
That was when I ran into the issue (merging 3.0 into our fork). I wrote this failing test on 3.0 as my base and then added in the nested processing to get it to pass. I'll look at why this is failing in CI (was working on my local build).

presidentbeef:
This is going to break so many fingerprints :sob:

ptoomey3:
oof...are you talking about these http://blog.presidentbeef.com/blog/2013/11/01/testing-brakeman/? That is quite unfortunate...sorry :frowning:

presidentbeef:
My testing is not really the issue, it's users who expect fingerprints to be relatively stable :( But it's okay, a necessary change.

presidentbeef:
Just as an FYI, working through bugs that are revealed by this change :crocodile:

ptoomey3:
:bow:

presidentbeef:
Thanks for your contribution and patience :ghost:

test/apps/rails4/lib/sweet_lib.rb
```diff
@@ -2,4 +2,9 @@ class SweetLib
   def do_some_cool_stuff bad
     `ls #{bad}`
   end
+
+  def test_find_group
+    #Should warn, no escaping done for :group
```
ptoomey3: Doh...just noticed this after the merge. I made a SQL based test on my firstbpass, but I changed it to be a command exec test and forgot to update the comment. My apologies.

test/apps/rails4/lib/sweet_lib.rb
```diff
@@ -2,4 +2,9 @@ class SweetLib
   def do_some_cool_stuff bad
     `ls #{bad}`
   end
+
+  def test_find_group
```
ptoomey3: Same with the test name. It should have been `test_remove` or some such thing.

test/apps/rails4/lib/sweet_lib.rb
presidentbeef: Does this make you feel better? https://github.com/presidentbeef/brakeman/pull/610

ptoomey3: :+1:

PR #606, opened 2015-01-06
add additional_libs option to process arbitrary additional directories

ptoomey3:
We had a patch to process the `lib` directory, as well as other directories that are "lib equivalent". Now that V3.0.0 processes the `lib` directory we thought it might be useful for upstream brakeman to support the same thing.

P.S. I will have to update the unit test after https://github.com/presidentbeef/brakeman/pull/605/ is merged, since the warning count will change.

/cc @gregose @mastahyeti @oreoshake

### DIFF
--- a/lib/brakeman/options.rb
+++ b/lib/brakeman/options.rb
@@ -124,6 +124,11 @@ module Brakeman::Options
           options[:skip_libs] = true
         end
 
+        opts.on "--add-libs-path path1,path2,etc", Array, "An application relative lib directory (ex. app/mailers) to process" do |paths|
+          options[:additional_libs_path] ||= Set.new
+          options[:additional_libs_path].merge paths
+        end
+
         opts.on "-t", "--test Check1,Check2,etc", Array, "Only run the specified checks" do |checks|
           checks.each_with_index do |s, index|
             if s[0,5] != "Check"
--- a/lib/brakeman.rb
+++ b/lib/brakeman.rb
@@ -17,6 +17,8 @@ module Brakeman
   #Options:
   #
   #  * :app_path - path to root of Rails app (required)
+  #  * :additional_checks_path - array of additional directories containing additional out-of-tree checks to run
+  #  * :additional_libs_path - array of additional application relative lib directories (ex. app/mailers) to process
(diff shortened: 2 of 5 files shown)


presidentbeef:
I think for less surprising behavior, paths should be relative to the current working directory, not relative to the root of the application being scanned.

Thoughts?

gregose:
> paths should be relative to the current working directory, not relative to the root of the application being scanned.

For our use case we are scanning application versions in various directories and keeping this path relative to the application allows us to have a standard config for each of these versions. If it was relative to the path running brakeman we would need to dynamically set this config for each scan.

lib/brakeman/options.rb
```diff
@@ -124,6 +124,11 @@ def get_options args, destructive = false
           options[:skip_libs] = true
         end
 
+        opts.on "--additional-libs path1,path2,etc", Array, "Process these additional lib directories" do |paths|
```
presidentbeef: For consistency with `--add-checks-path` can we call this `--add-libs-path`?

presidentbeef:
Hm, I see your point. Alternatively, you could run your scans from the root of the application.

I am concerned about this, because it is inconsistent with the way `--add-checks-path` works and it prevents pointing the path to anything outside of the application to be scanned.

On the other hand, I imagine very few people will even notice.

presidentbeef:
In any case, please merge/rebase master as mentioned above.

lib/brakeman/options.rb
```diff
@@ -124,6 +124,11 @@ def get_options args, destructive = false
           options[:skip_libs] = true
         end
 
+        opts.on "--add-libs-path path1,path2,etc", Array, "An application relative lib directory (ex. app/mailers) to process" do |paths|
```
ptoomey3: Ok, I added options descriptions for the new `additional_libs_path` as well as the previously undocumented `additional_checks_path` above. I've tried to clear up the ambiguity between these path types by more clearing defining what each is for. Hopefully the "application relative lib directory" language alon with the example "app/mailers" will make it clear what this option is for.

presidentbeef:
Works for me. Thanks! :rocket:

gregose:
:metal:

PR #607, opened 2015-01-06
allows non-standard directory structure

nealharris:
The motivation for this is that I work with a Rails app whose maintenance is shared by many small teams.  We organize the code such that teams have their own sub-dirs under `app`.  So, we have `app/foo_team/controllers/api/foo_controller.rb`, and `app/bar_team/models/bar.rb`, etc.

This change allows me to continue to use Brakeman with the code organized this way, but shouldn't have any adverse affects for apps that have the usual Rails structure.


lib/brakeman/version.rb
```diff
@@ -1,3 +1,3 @@
 module Brakeman
-  Version = "3.0.0"
```
presidentbeef: Please do not change the version number.

nealharris: Done.  Sorry about that.

test/tests/brakeman.rb
```diff
@@ -29,6 +29,42 @@ def test_relative_path_in_warnings
 
 
   end
+
+  def test_app_tree_flexible_file_paths
+    require 'brakeman/options'
+    relative_path = File.expand_path(File.join(File.dirname(__FILE__), "..", "/apps/rails4_non_standard_structure/"))
```
presidentbeef: You can use `TEST_PATH` here:

```ruby
File.join(TEST_PATH, "/apps/rails4_non_standard_structure/")
```

test/tests/brakeman.rb
```diff
@@ -29,6 +29,42 @@ def test_relative_path_in_warnings
...
+    at.controller_paths.each do |path|
+      if File.basename(path) == 'foo.rb'
+        contains_foo_controller = true
+        break
+      end
+    end
+
+    contains_foo_model = false
+    at.model_paths.each do |path|
+      if File.basename(path) == 'foo.rb'
```
presidentbeef: These file names should probably be different so it is clear the test is working correctly.

nealharris: Agree.  Done.

test/tests/brakeman.rb
nealharris: Done.

presidentbeef:
Doesn't seem to break anything.

Thanks!

PR #611, opened 2015-01-22
Do not warn about missing CSRF protection in API

### LINKED ISSUE #573: protect_from_forgery and Rails API
Hi, `protect_from_forgery` do not apply to Rails API but `brakeman` can't tell the difference and raise warning.

presidentbeef:
If ApplicationController does not inherit from ActionController::Base, do not warn about missing/disabled `protect_from_forgery`.

This would close #573.


futhr:
:+1:

PR #626, opened 2015-02-24
Better render path information for templates

presidentbeef:
Use an actual data structure instead of an array of strings with semantics embedded in them.

Ideally this information should then be available in the JSON report, but that is a pretty serious change so for now it's only an internal change.


oreoshake:
ooOOOOOoohhhhh :+1: :mortar_board:

PR #634, opened 2015-03-09
Don't ignore calls in models outside methods

presidentbeef:
For whatever reason, calls outside of methods were ignored after initial processing, and did not go through alias processing. This meant constants were not propagated and things like [this happened](http://stackoverflow.com/questions/27853371/rails-brakeman-warning-of-sql-injection) when block/lambda arguments were not processed.


presidentbeef:
Also updated how default arguments are handled because using the default value can cause issues to be hidden.

PR #637, opened 2015-03-18
add rake task to exit with error code when fail

masarakki:
it's useful for CI like travis.


presidentbeef:
Hi,

Thank you for this. I'm sure someone will find it useful, although I discourage using `rake` with Brakeman since it unnecessarily loads the entire Rails application.

tunacasserole:
`result.filtered_warnings.empty? was clutch.  I can use that to send an email if any warnings are found without having to parse the output file.

PR #648, opened 2015-04-14
Warn unless a rails 4 app protect[s]_from_forgery with :exception

oreoshake:
Because this has bitten me twice before and the effect is basically catastrophic, let's warn people when they don't use `protect_from_forgery with: :exception`. This is now the default for newly generated rails 4 applications and so it should be considered the standard.

### DIFF
--- a/lib/brakeman/checks/check_forgery_setting.rb
+++ b/lib/brakeman/checks/check_forgery_setting.rb
@@ -54,14 +54,22 @@ class Brakeman::CheckForgerySetting < Brakeman::BaseCheck
         :link_path => "https://groups.google.com/d/topic/rubyonrails-security/LZWjzCPgNmU/discussion"
     elsif version_between? "4.0.0", "100.0.0" and forgery_opts = app_controller[:options][:protect_from_forgery]
 
-      unless forgery_opts.is_a?(Array) and sexp?(forgery_opts.first) and hash_access(forgery_opts.first.first_arg, :with).value == :exception
-        warn :controller => :ApplicationController,
+      unless forgery_opts.is_a?(Array) and sexp?(forgery_opts.first) and
+          access_arg = hash_access(forgery_opts.first.first_arg, :with) and access_arg.value == :exception
+
+        args = {
+          :controller => :ApplicationController,
           :warning_type => "Cross-Site Request Forgery",
           :warning_code => :csrf_not_protected_by_raising_exception,
           :message => "protect_from_forgery should be configured with 'with: :exception'",
           :confidence => CONFIDENCE[:med],
           :file => app_controller[:files].first,
           :link_path => "blog post link?"
+        }
+
+        args.merge!(:code => forgery_opts.first) if forgery_opts.is_a?(Array)
+
+        warn args
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/brakeman/checks/check_forgery_setting.rb
```diff
@@ -52,6 +52,17 @@ def run_check
         :confidence => CONFIDENCE[:high],
         :gem_info => gemfile_or_environment,
         :link_path => "https://groups.google.com/d/topic/rubyonrails-security/LZWjzCPgNmU/discussion"
+    elsif version_between? "4.0.0", "100.0.0" and forgery_opts = app_controller[:options][:protect_from_forgery]
+
+      unless forgery_opts.is_a?(Array) and sexp?(forgery_opts.first) and hash_access(forgery_opts.first.first_arg, :with).value == :exception
```
presidentbeef: It's possible `hash_access(forgery_opts.first.first_arg, :with)` will return `nil`.

lib/brakeman/checks/check_forgery_setting.rb
```diff
@@ -52,6 +52,17 @@ def run_check
...
+        warn :controller => :ApplicationController,
+          :warning_type => "Cross-Site Request Forgery",
+          :warning_code => :csrf_not_protected_by_raising_exception,
+          :message => "protect_from_forgery should be configured with 'with: :exception'",
+          :confidence => CONFIDENCE[:med],
+          :file => app_controller[:files].first,
+          :link_path => "blog post link?"
```
presidentbeef: Or update http://brakemanscanner.org/docs/warning_types/cross-site_request_forgery/ and link to external blog posts from there. Then just remove `:link_path`.

lib/brakeman/checks/check_forgery_setting.rb
```diff
@@ -52,6 +52,17 @@ def run_check
         :confidence => CONFIDENCE[:high],
         :gem_info => gemfile_or_environment,
         :link_path => "https://groups.google.com/d/topic/rubyonrails-security/LZWjzCPgNmU/discussion"
+    elsif version_between? "4.0.0", "100.0.0" and forgery_opts = app_controller[:options][:protect_from_forgery]
+
+      unless forgery_opts.is_a?(Array) and sexp?(forgery_opts.first) and hash_access(forgery_opts.first.first_arg, :with).value == :exception
+        warn :controller => :ApplicationController,
```
presidentbeef: Pass in `forgery_opts.first` as `:code`? Then we'd also get the line number.

lib/brakeman/warning_codes.rb
```diff
@@ -87,6 +87,7 @@ module Brakeman::WarningCodes
     :CVE_2011_2932 => 83,
     :cross_site_scripting_inline => 84,
     :CVE_2014_7829 => 85,
+    :csrf_not_protected_by_raising_exception => 86,
```
presidentbeef: Thanks for adding this!

lib/brakeman/checks/check_forgery_setting.rb
oreoshake: done https://github.com/presidentbeef/brakeman-site/pull/17

lib/brakeman/checks/check_forgery_setting.rb
oreoshake: fixed in 12ed187aaa85ab560c927bc8585f75e3a6efefbf

lib/brakeman/checks/check_forgery_setting.rb
oreoshake: done 12ed187aaa85ab560c927bc8585f75e3a6efefbf. sometimes the value is a boolean so there's some code to handle that.

PR #651, opened 2015-04-22
Only run scans when test runs

presidentbeef:
Also, since scans are no longer run upfront, make them quiet so they don't mess up the test output.

/cc @ichthala

### DIFF
--- a/test/test.rb
+++ b/test/test.rb
@@ -25,19 +25,9 @@ module BrakemanTester
     #Run scan on app at the given path
     def run_scan path, name = nil, opts = {}
       opts.merge! :app_path => "#{TEST_PATH}/apps/#{path}",
-        :quiet => false,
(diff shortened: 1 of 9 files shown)


ichthala:
It looks like the rake task won't run these tests anymore, since run_scan will no longer be called when the files are required. Is there an alternate way to run them that I'm not seeing?

presidentbeef:
Hm, running with rake seems to work fine for me and Travis?

Alternatively you can run `ruby test/test.rb` (which is exactly what the rake task does).

ichthala:
I see what I was missing — I didn't realize that the tests sent the `report` message themselves, so I thought that `run_scan` would only be called when the files were required in test.rb.

Also, it looks like the tests for scans with the :only_files option weren't changed over to the new pattern. Is that intentional?

presidentbeef:
Nope, I missed that one, thanks! Updated.

PR #658, opened 2015-05-21
load rake tasks

masarakki:
on rails project, load brakeman tasks automatically.


presidentbeef:
Hi Masarakki,

Thank you for your effort.

However, after some consideration, I have decided not to accept this change. It encourages several behaviors which I am trying to discourage:

* Running Brakeman from a `rake` task
* Adding Brakeman to `Gemfile`
* Not adding `require: false` in `Gemfile`

I hope you do not mind. Apologies for the delay.

PR #662, opened 2015-06-08
Add Support for a progress dot formatter similar to rspec's default format

anfleene:
Thanks for making such an awesome Gem! 

I'd like to be able to track progress similar to the rspec dot formatter. This seemed like a pretty straightforward way of achieving that. I turned it off by default leaving the in place progress formatter as the default.

I also added a bit more whitespace and removed the trailing dots/whitespace from each section. This seemed a bit more clear even with the default progress format.

### DIFF
--- a/lib/brakeman.rb
+++ b/lib/brakeman.rb
@@ -134,6 +134,7 @@ module Brakeman
       :parallel_checks => true,
       :relative_path => false,
       :report_progress => true,
+      :progress_dots => false,
       :html_style => "#{File.expand_path(File.dirname(__FILE__))}/brakeman/format/style.css"
     }
   end
--- a/lib/brakeman/options.rb
+++ b/lib/brakeman/options.rb
@@ -31,6 +31,10 @@ module Brakeman::Options
           options[:report_progress] = progress
         end
 
+        opts.on "--[no-]progress-dots", "Show progress as dots" do |progress_dots|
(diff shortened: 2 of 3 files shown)


presidentbeef:
Hi Andy,

Thanks for this. However, I can't say I prefer this output formatting to the current version.

For example:

```
Loading scanner...
Processing application in /Users/justinc/scratch/brakeman/test/apps/redmine
Processing gems
[Notice] Detected Rails 4 application

Processing configuration
[Notice] Escaping HTML by default

Parsing files

Processing initializers
 3/4 files processed
Processing libs
 126/127 files processed
Processing routes

Processing templates
 328/329 templates processed
Processing data flow in templates
 316/317 templates processed
Processing models
 71/72 files processed
Processing controllers
 47/48 files processed
Processing data flow in controllers
 47/48 controllers processed
Indexing call sites
```

The gaps are a little weird, and it's confusing that it's always `n - 1` things processed. If this change were to stay, then I'd want to change when the progress is reported (i.e. after processing instead of before).

While the dots look cool, I don't see a lot of value in them over the `x/total` progress reporting. I don't imagine a lot of people would use this option just for the cosmetic difference (especially when they often don't notice all the functional options...)

Sorry, but I don't think I'll be accepting this change.

anfleene:
No worries. Thanks for the consideration.

PR #663, opened 2015-06-10
Handle <%== as raw output in ERB templates

presidentbeef:
`<%== whatever %>` should result in unescaped/raw output.

Sadly, I fixed parsing double equals in 46b97150eb05eb4425813812927a246e87a5b3c0 but didn't add a test for XSS.


cktricky:
Checking

cktricky:
Didn't flag AFAICT, I created a branch for you:

https://github.com/OWASP/railsgoat/tree/brakeman_test

https://github.com/OWASP/railsgoat/blob/brakeman_test/app/views/layouts/shared/_header.html.erb#L29

presidentbeef:
In that case it's because Brakeman doesn't know what `current_user` is. If you change it to `@current_user` it will get flagged. :/

PR #673, opened 2015-06-27
Do not warn about hrefs with *_path method

presidentbeef:
@oreoshake ?


presidentbeef:
Actually, never mind. I have a better change coming that will also fix this.

PR #674, opened 2015-06-28
Several XSS check fixes

presidentbeef:
* `safe_methods` and `url_safe_methods` will match any method, even if they have a target
* Fix how ignored methods were handled in XSS checks (they weren't being checked in immediate output)
* Ignore `*_path` methods as URLs
* Ignore `polymorphic_url` as URL
* Ignore `*_url` methods as URLs but not if they have a target
* Warning code for low confidence XSS warnings was for JSON output (`5`) instead of plain XSS (`2`)


presidentbeef:
Tests coming up

PR #675, opened 2015-06-30
Treat String#html_safe the same as raw()

presidentbeef:
That is, as unescaped output.

I believe previously the reasoning was `html_safe` was applied on purpose and so we shouldn't warn. Given how often `html_safe` confuses people, maybe that is not a valid assumption.


oreoshake:
:heart_eyes_cat: :heart_eyes_cat: :heart_eyes_cat:

rbhitchcock:
I think this would be a good, useful change. I frequently see it used incorrectly when evaluating applications. Perhaps an option to disable this behavior might be nice, but I don't think that would be necessary by any stretch.

presidentbeef:
Didn't think it would be controversial, but thought I'd check anyway :panda_face:

PR #676, opened 2015-07-02
Remove unnecessary new call creation in template processing

presidentbeef:


presidentbeef:
Wow that "TODO" seemed really recent to me, but it's from 2012.

PR #677, opened 2015-07-02
Relax dependency on highline

johnnyshields:
It conflicts with Capistrano and others.

### DIFF
--- a/brakeman.gemspec
+++ b/brakeman.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.add_dependency "ruby2ruby", "~>2.1.1"
   s.add_dependency "terminal-table", "~>1.4.5"
   s.add_dependency "fastercsv", "~>1.5"
-  s.add_dependency "highline", "~>1.6.20"
+  s.add_dependency "highline", ">=1.6.20"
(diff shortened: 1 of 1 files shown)


presidentbeef:
What are "and others"? Capistrano doesn't depend directly on highline. If you have the output from bundler, that would be useful. 

Highline 1.7.0 dropped Ruby 1.8 support but I'm not ready to give up on it. If you really need the latest highline, you can use brakeman-min.

johnnyshields:
So, it looks like I can safely downgrade highline, but why not relax the dependency on highline within Brakeman itself and just add a note "If you are using Ruby 1.8, please set `highline <= 1.6`". This way users can choose.

(For CI purposes you can setup Travis so the Ruby 1.8 build uses a 1.8-specific gemfile which has highline 1.6)

On a separate note, anyone who is serious enough about software security to use Brakeman should stop using Ruby 1.8!

presidentbeef:
This still sounds like a Bundler problem to me if you were able to resolve the dependency manually.

I'll think about options for the next major release.

presidentbeef:
This will updated in the next release. Thanks!

johnnyshields:
@presidentbeef thanks for helping out

Issue #430, opened 2014-01-21
XML report?


abedra:
I know there are plenty of report formats already, but would you object to XML reporting as an option? I will happily do the work.

presidentbeef:
JUnit XML was added in #1453 👍

PR #679, opened 2015-07-08
Added XML reporting #430

jacarandang:
I tried working on XML report. I'm really new to brakeman and XML, any comments would be highly appreciated.
#430


presidentbeef:
Hi,

Thanks for looking into this!

`:xml` should be added as an option here: https://github.com/presidentbeef/brakeman/blob/master/lib/brakeman/options.rb#L166-L169

Unfortunately, your branch isn't working for me on any apps I've tested. This is the error:

```
/home/justin/.rvm/gems/ruby-2.1.5@brakeman/gems/builder-3.2.2/lib/builder/xmlbase.rb:61:in `tag!': XmlMarkup cannot mix a text argument with a block (ArgumentError)
	from /home/justin/.rvm/gems/ruby-2.1.5@brakeman/gems/builder-3.2.2/lib/builder/xmlbase.rb:93:in `method_missing'
	from /home/justin/work/brakeman/lib/brakeman/report/report_xml.rb:48:in `output_xml'
	from /home/justin/work/brakeman/lib/brakeman/report/report_xml.rb:33:in `generate_report'
	from /home/justin/work/brakeman/lib/brakeman/report.rb:76:in `generate'
	from /home/justin/work/brakeman/lib/brakeman/report.rb:45:in `format'
	from /home/justin/work/brakeman/lib/brakeman.rb:356:in `block (2 levels) in write_report_to_files'
	from /home/justin/work/brakeman/lib/brakeman.rb:355:in `open'
	from /home/justin/work/brakeman/lib/brakeman.rb:355:in `block in write_report_to_files'
	from /home/justin/work/brakeman/lib/brakeman.rb:354:in `each'
	from /home/justin/work/brakeman/lib/brakeman.rb:354:in `each_with_index'
	from /home/justin/work/brakeman/lib/brakeman.rb:354:in `write_report_to_files'
	from /home/justin/work/brakeman/lib/brakeman.rb:343:in `scan'
	from /home/justin/work/brakeman/lib/brakeman.rb:62:in `run'
	from /home/justin/work/brakeman/bin/brakeman:78:in `<main>'
```

jacarandang:
Thank you for checking my PR. Can I see the text output? I think it is caused by errors.

jacarandang:
As expected. the errors was caused by the error reports. Can you try again now? I'll add tests later. Thank you

presidentbeef:
I do not think errors are the problem. Pick any of the apps in https://github.com/presidentbeef/brakeman/tree/master/test/apps (none of which should have errors) and you will still see the issue.

jacarandang:
Weird, I tried it against the apps and it did work well. I'll check on this, I'll try to make tests. Thanks

presidentbeef:
I'm using Ruby 2.1.5 with Builder 3.2.2, if that helps.

jacarandang:
I think the send method is causing the problem. It might also be overwritten by Builder, I used the tag! method instead and added sanity check.

jacarandang:
Builder doesn't seem to support ruby 1.8.7 which causes the travis to fail.

lib/brakeman/report/report_xml.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    output_xml x, :report_info, reportInfo
+
+  end
+
+  def convert_to_hashes warnings, typ
+    warnings.map do |w|
+      hash = w.to_hash
+      hash[:file] = warning_file w
+      hash[:location] = hash[:location].flatten.join(',') unless hash[:location].nil?
```
presidentbeef: Shouldn't this be nice XML, not just a string?

lib/brakeman/report/report_xml.rb
```diff
@@ -0,0 +1,71 @@
...
+    scanInfo = {
+      :app_path => File.expand_path(tracker.app_path),
+      :ruby_version => RUBY_VERSION,
+      :rails_version => rails_version,
+      :brakeman_version => Brakeman::Version,
+      :report_generation_time => Time.now.to_s,
+      :start_time => tracker.start_time.to_s,
+      :end_time => tracker.end_time.to_s,
+      :duration => tracker.duration,
+      :checks => checks.checks_run.sort.join(',')
```
presidentbeef: Shouldn't this be a list in XML, not a string?

jacarandang: What do you mean? Is this the checks?

lib/brakeman/report/report_xml.rb
jacarandang: Is this about the location?

lib/brakeman/report/report_xml.rb
presidentbeef: Yes. Currently it is turning an array into a string.

lib/brakeman/report/report_xml.rb
presidentbeef: Yes.

presidentbeef:
It would be nice if the test could parse the XML back into a hash and compare it with the expected values.

brakeman-min.gemspec
```diff
@@ -18,4 +18,5 @@ Gem::Specification.new do |s|
   s.add_dependency "ruby_parser", "~>3.7.0"
   s.add_dependency "ruby2ruby", "~>2.1.1"
   s.add_dependency "multi_json", "~>1.2"
+  s.add_dependency "builder", "~>2.0"
```
presidentbeef: Don't add builder here, brakeman-min only supports JSON by default.

lib/brakeman/report/report_xml.rb
```diff
@@ -0,0 +1,78 @@
+require 'builder'
```
presidentbeef: Please use `Brakeman.load_brakeman_dependency 'builder'` instead of `require`

presidentbeef:
I'm not sure we should use `builder` given it seems unmaintained since Jim Weirich passed away, but it's probably okay.

jacarandang:
How about using Nokogiri instead? Could also help in parsing XML files for tests.

presidentbeef:
That would be fine with me.

dvandersluis:
FYI - I've been doing a bunch of testing of XML builders lately, and the `builder` gem is quite slow. `nokogiri` is significantly faster, and in my testing [`ox`](https://github.com/ohler55/ox) is even better (it's recently been updated to have a module for building XML documents).

presidentbeef:
I'm open to another attempt at this with a maintained library.

PR #684, opened 2015-07-16
Update Slim, Highline, Ruby2Ruby dependencies

presidentbeef:


bf4:
Now everyone using brakeman must install *both slim and haml*, even if their app uses neither?

presidentbeef:
Yes, this has always been the case so that Brakeman works out of the box for scanning any application. Brakeman was never intended to be included in Gemfiles, it is a standalone application. If you don't like having extra dependencies, you can use the `brakeman-min` gem and install what you need.

bf4:
ah, good call

PR #685, opened 2015-07-16
Use highline 1.x series >= 1.6.20 rather than 1.6.x

bf4:
ref: http://www.benjaminfleischer.com/2013/08/21/psa-gem-versions/

Without this change, brakeman is incompatible with the libraries using the 1.7 series of highline, which there doesn't appear to a reason for.


presidentbeef:
The reason was Ruby 1.8, see https://github.com/presidentbeef/brakeman/pull/677,  https://github.com/presidentbeef/brakeman/pull/684, and https://twitter.com/brakeman/status/621337183253729280

This will be fixed in #684.

bf4:
Ha!  I just looked at the history of the gemspec, not for other issues. Thanks!

Issue #686, opened 2015-07-20
Brakeman 3.0.4, 3.0.5 stuck


dfens:
After upgrading brakeman to version 3.0.4 or 3.0.5 brakeman is totally stuck:
```
brakeman --format=html -d -o brakeman-report.html
Loading scanner...
Processing application in /home/...
Processing gems...
[Notice] Detected Rails 3 application
Processing configuration...
[Notice] Escaping HTML by default
Parsing files...
```

And this is all. No 'processing' no anything just waiting loop. Any toughts? I saw ruby parser was updated in 3.0.4

presidentbeef:
Hi Paweł,

Sorry for the trouble. Can you check out the `report_file_being_parsed` branch and determine which file is the problem? Then maybe we can get it fixed in ruby_parser. Thanks!

dfens:
Thanks,

it seems that it is stuck in file with inline style base64 images:
```
      .tutorial_content .shadow {
        position: absolute;
        background-image: url(data:image/png;base64,iVBORw..... kilobytes of data...
```

presidentbeef:
Can you provide a more complete snippet? Is it HAML, ERB, etc? I don't need the image itself, but how the CSS is embedded in the file.

dfens:
this is html erb,
please find following gist:
https://gist.github.com/dfens/ee9d8552b868018d7027

also there is jquery included inline which I missed yesterday.
I put there only <head> content, I don't think that body does matter

presidentbeef:
Thanks, I am fairly certain the giant jquery string was the problem.

This should be fixed with https://github.com/seattlerb/ruby_parser/pull/197

zenspider:
I'm confused. Why is an html file being parsed by ruby_parser? I see no ERB tags in there at all.

presidentbeef:
Closing since this is fixed in ruby_parser.

presidentbeef:
Please update to ruby_parser 3.7.1 and see if that resolves the issue.

PR #687, opened 2015-07-20
Downgrade ruby_parser to 3.6.6 since 3.7.0 gets stuck

dfens:
In reference to #686 - downgrading ruby_parser fixes problem


presidentbeef:
Not going to go backwards, this will be fixed in ruby_parser itself.

PR #690, opened 2015-07-24
Add Model, Controller, Template, Config classes

presidentbeef:
Instead of using hashes for everything.


oreoshake:
dang :rocket: :boom:

PR #695, opened 2015-08-05
Fallback to the `railties` version when the `rails` gem is missing.

lucasmazza:
Rails apps that don't include `actionmailer` or `activerecord` might not have
the meta `rails` gem on their Gemfile, but the `railties` gem instead.

I'm not sure of where to add a test for this, and if makes sense to duplicate some of the Rails apps used for testing just to test this case, but if it makes sense I can duplicate the `rails4` app as a `rails4-railtie` one.

### DIFF
--- a/lib/brakeman/tracker/config.rb
+++ b/lib/brakeman/tracker/config.rb
@@ -62,7 +62,7 @@ module Brakeman
 
     def set_rails_version
       # Ignore ~>, etc. when using values from Gemfile
-      version = gem_version(:rails)
+      version = gem_version(:rails) || gem_version(:railties)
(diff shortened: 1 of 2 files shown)


presidentbeef:
Hi Lucas,

Thank you for adding this! You should be able to add a test pretty easily to https://github.com/presidentbeef/brakeman/blob/master/test/tests/cves.rb Let me know if you need help!

lucasmazza:
@presidentbeef just re-pushed with a small test using the same pattern as the CVEs tests, thanks for the tip.

presidentbeef:
Thanks!

PR #696, opened 2015-08-07
Convert any string config keys into symbols.

KarlGl:
After parsing the yaml config file for options, convert any string config keys into symbols.
I had to add this to get the following config file working:
```
skip_checks:
  - CheckMassAssignment
```
Otherwise the default options `:skip_checks` key takes precedence.

Am I doing something wrong? I couldn't find examples of how this config file should look like.


presidentbeef:
Hi Karl,

It's generally expected that people will use `-C` to generate the configuration files instead of needing to do it by hand. That said, this may still be a good change to make it easier to use.

PR #703, opened 2015-08-23
Rubocop

cheerfulstoic:
Hello!

So I don't know if you have heard of rubocop, but I figured you'd be open to use a static code analysis tool ;)

I heard you on Ruby Rogues and when I saw your repo I noticed you had a 2.0 on score in Code Climate.  You also said that you've been doing a lot of refactoring since you've been writing the code for a long time and have improved as a programmer.  Rubocop is a great way to help you quickly find places in your code for improvement and, in some cases, it will fix those things for you!  Also, as you said, tools like this can be good learning tools for when something is pointed out to you which causes you to do research on it.

Quick summary: rubocop is a tool to statically analyze your code to find issues with code style, performance, and code complexity.  It's configured with the `.rubocop.yml` file which defines how the various "cops" should behave.

What I've done in this pull request is to set up rubocop with the `rubocop --auto-gen-config` command.  This creates a `.rubocop_todo.yml` file which can be included in your `.rubocop.yml`.  The `--auto-gen-config` argument finds all of the "problems" (according to rubocop's default idea of what a problem is) and creates the todo file which tells rubocop to ignore all of those problems.  That way you can, at your leisure, remove items from the todo file in order to take care of them (or, in some cases, to put the ignore into the `.rubocop.yml` file instead to say that you want your project to work in a certain way).  Another nice thing about this is that once you've enabled a cop (or in some cases lowered the metric's max value) then rubocop can help check that future code doesn't get any worse.

In addition to creating the todo file I've also removed and had rubocop automatically fix a number of the cops so that you can get an idea for some of what rubocop does.  For each cop I've made a separate commit so that it's easy for you to soo what each cop does and even `git revert` if you don't think that cop  is a good idea.

We use rubocop in the neo4j gem and it's been a really nice way to clean things up and improve our Code Climate score.  The Code Climate will mainly come from the `Metrics/*` cops which analyze code complexity.  Here is our `.rubocop.yml` file:

https://github.com/neo4jrb/neo4j/blob/master/.rubocop.yml

You'll note that we've used it to configure a few things to our taste.  Also our todo file still has a number of items in in.  We've been using rubocop for 6 months and so it takes some time to work through all of the issues.  But I find it very satisfying ;)

I've also enabled rubocop to run in Travis CI, so hopefully this PR will run it and you can see it in action.  There are a couple of Ruby syntax failures that I didn't try to correct in `test/apps/rails2/lib/generators/test_generator/templates/model.rb`.  I think if you were to rename the file with a `.erb` extension rubocop would be happy.

I'm happy to answer questions!


presidentbeef:
Hi Brian,

I think Rubocop is a fine tool and I appreciate your work here. However, I don't think these changes are particularly helpful in making Brakeman better.

Thanks, though!

cheerfulstoic:
I agree that these aren't groundbreaking changes.  The reason for that is because I didn't want to introduce anything that I thought that you would disagree with.  The point of the PR is to introduce Rubocop into the project so that you can continuously improve breakman in the future.  The biggest place that I see that happening is with the code metrics cops which help you find the complexity in your code.

PR #706, opened 2015-08-25
Allow support for j and escape_javascript

kbrock:
Hello,

The `j` and `escape_javascript` method are used in javascript blocks to ensure that
text is properly escaped to be in a javascript string.

This works much the same way as `escapeHTML` and `h` do for ensuring that text is safe for inclusion in an html tag.

This can be found throughout stack overflow and the docs can be found [here](http://apidock.com/rails/ActionView/Helpers/JavaScriptHelper/escape_javascript).

Does this sounds like a good inclusion?
What else can I provide to help drive that decision?

Thanks for this great tool

### DIFF
--- a/lib/brakeman/checks/check_cross_site_scripting.rb
+++ b/lib/brakeman/checks/check_cross_site_scripting.rb
@@ -287,9 +287,9 @@ class Brakeman::CheckCrossSiteScripting < Brakeman::BaseCheck
   end
 
   def setup
-    @ignore_methods = Set[:button_to, :check_box, :content_tag, :escapeHTML, :escape_once,
+    @ignore_methods = Set[:button_to, :check_box, :content_tag, :escapeHTML, :escape_javascript, :escape_once,
                            :field_field, :fields_for, :h, :hidden_field,
-                           :hidden_field, :hidden_field_tag, :image_tag, :label,
+                           :hidden_field, :hidden_field_tag, :image_tag, :j, :label,
                            :link_to, :mail_to, :radio_button, :select,
                            :submit_tag, :text_area, :text_field,
                            :text_field_tag, :url_encode, :url_for,
--- a/test/apps/rails_with_xss_plugin/app/views/layouts/users.html.erb
+++ b/test/apps/rails_with_xss_plugin/app/views/layouts/users.html.erb
@@ -6,6 +6,9 @@
   <meta http-equiv="content-type" content="text/html;charset=UTF-8" />
   <title>Users: <%= controller.action_name %></title>
   <%= stylesheet_link_tag 'scaffold' %>
+<script>
+  var browser="<%= j request.env['HTTP_USER_AGENT'] %>";
+</script>
 </head>
 <body>
 


oreoshake:
@kbrock the `escape_javascript` method is only safe when in javascript contexts. Outside, it can still lead to xss. e.g. `escape_javascript("")` does nothing to stop the xss.

oreoshake:
Hmm well `escape_javascript` doesn't mark it as `html_safe` so the content will still be escaped by default.

kbrock:
@oreoshake - thanks

`j` is used to escape a javascript string literal. If the user inserts that directly into the dom, or uses it outside that scope, then you are right, it doesn't work.
But on that same note, `h` is used to insert into html. It is not made for a javascript string literal.

```erb
<!-- incorrect use of j -->
<script><%= j param[:x] %></script> <!-- 1: bad: alert('x') -->
<textarea><%=j param[:x] %></textarea><!-- 2: bad: </textarea><script>alert('x');</script><textarea> -->
<!-- correct use of j -->
<script>alert("<%= j param[:x] %>");</script> <!-- 3: safe -->
<!-- incorrect use of h -->
<script>var j = "<%= h param[:x] %>";</script> <!-- 4: unsafe: ";alert('x');" -->
```

From what I've read, using `j` is a good practice. My goal is to remove the false positive around case 3. 

Options:

1. Warn for `j` and not `h`, and encourage developers to always use `h`, even when vulnerable.
2. Assume that developers are using `j`, they are using it properly, and are using `h` properly for that matter.
3. Make brakeman smart enough to detect when `erb`, `haml`, `slim`, ... jump from html to javascript via `javascript_tag`, `<script>`, `onclick`). And ensure the `j` is surrounded by double quotes.

Current implementation is option 1, this PR is option 2. Option 3 looks great, but seems daunting.

Thanks,
Keenan

kbrock:
Ok, that was mostly false.

It looks like main difference between `j` and `h` is when the end of the parameter is a `\`.

```erb
<script>var h = "<%= h param[:x] %>";</script> <!-- bad: abc\ -->
```

presidentbeef:
@kbrock why did you add a test case to an old Rails 2 app using the `rails_xss` plugin? Seems like an odd choice. The output of `<%=` would be escaped anyhow.

It looks like to me `j` is not as safe in Rails 2.x, since the characters which are escaped have been expanded since then and in Rails 3 and 4 the output of `j` will be html escaped by default anyway. This suggests the only time Brakeman would warn about `j` would be in Rails 2 apps or newer apps where the output of `j` is explicitly not escaped (with `raw` or `html_safe`).

What's the scenario where you are seeing a warning?

kbrock:
@presidentbeef Thanks for the feedback. I am not sure the best way to add this to the code.

We were getting some warning in rails 4.2.x haml views. So we put `j` into there to properly escape the javascript, but it still complained. When I used `h` instead, it liked it. So I was trying to fix brakeman so it properly work when using the correct method, `j`.

[code](https://github.com/ManageIQ/manageiq/pull/4016/files#diff-92237fc7f1a886c82c4758d7b315a979R42):

```haml
  :javascript
    var import_file_upload_id = "#{j(@import_file_upload_id)}";
```

running [brakeman on travis](https://travis-ci.org/ManageIQ/manageiq/jobs/76995654) it raising an XSS warning:

```
Weak
miq_policy/import (MiqPolicyController#import)
Cross Site Scripting
Unescaped parameter value near line 42:
find_and_preserve(Haml::Filters::Javascript.render_with_options("var import_file_upload_
```

presidentbeef:
Ah, okay. This is actually something different. Essentially I need to work a little more on the HAML handling so that `find_and_preserve` is not treated like a normal call. It's on my list of things to do...

kbrock:
Thanks so much @presidentbeef 

Using `--safe-methods j,escape_javascript -C` is great and is working for us in the short term. Was hoping to give a little back here, but sounds like it is quite an undertaking. never enough time.

Do you want to close this? I could create a separate issue instead.

presidentbeef:
Sounds good.

PR #710, opened 2015-09-01
Allow method to be specified by regex

aianus:
The options to `find_calls` includes `:method - symbol, array of symbols, or regular expression to match method(s)` but passing a regular expression without a target does not actually work without this patch.

### DIFF
--- a/lib/brakeman/call_index.rb
+++ b/lib/brakeman/call_index.rb
@@ -139,6 +139,8 @@ class Brakeman::CallIndex
   def calls_by_method method
     if method.is_a? Array
       calls_by_methods method
+    elsif method.is_a? Regexp
+      calls_by_methods_regex method
     else
       @calls_by_method[method.to_sym] || []
     end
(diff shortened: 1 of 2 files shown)


presidentbeef:
Hi Alex,

Yes, it was removed for performance reasons: http://blog.presidentbeef.com/blog/2012/11/28/faster-call-indexing-in-brakeman-1-dot-8-3/

Do you need this functionality? If so we'll have to look into better performing code. Right now it would convert every single call to a string every time it tried to match a regex. This could make it even slower than it was before.

aianus:
Yes, we are using this functionality.

Seems to me it should be up to the author and user of the check if they want to use this slower search method. For our use case at Coinbase searching by regex was plenty fast and our app is pretty 'large'.

presidentbeef:
It's possible to support it and I'm happy to do so if it's something you are using. The trouble previously was converting the target to strings, while this change is only doing methods which is considerably cheaper.

PR #714, opened 2015-09-11
Calculate absolute paths correctly for Windows

spcoder:
The current method of determining if a file is absolute does not work
correctly on Windows. The result is that ignore files contain an
absolute path instead of a relative path. This also leads to an
incorrect fingerprint which prevents developers and continuous
integration tools from sharing ignored warnings.


presidentbeef:
Thanks!

PR #722, opened 2015-09-21
Add optional check for weak hashing algorithms (MD5/SHA1)

presidentbeef:
If the digest contains user input or variables that might be passwords, the confidence is high. Otherwise medium.


oreoshake:
Will this catch `OpenSSL::Digest::SHA1/MD5`?

presidentbeef:
Mmm no.

presidentbeef:
Added in #724

PR #726, opened 2015-09-26
Sortable HTML report tables

dlanner:
This PR makes all the tables in the HTML report sortable. If you click on a column header, it will sort the table by that column. Clicking it after that toggles the order of the sort.

This is useful for when you want to group together all security warnings by Class or Warning Type, for example.

For the sortable functionality, I used Data Tables, which depends on jQuery. In order to keep the report to a single file that can be transferred around easily, the javascript files are downloaded from CDNs (CloudFlare for Data Tables, MaxCDN for jQuery) using HTTPS. I used the minified versions of both scripts to minimize page load times.

If the report viewer's web browser is setup to block either the jQuery or Data Tables download requests for whatever reason, the report will be unaffected; it will look unchanged, only there will be a couple error messages in the javascript console. But someone whose browser is setup to block jQuery would be used to those console errors I'm sure, given the library's ubiquity.

A summary of the 3 main changes in this PR:
* Initializing the Data Tables: $('table').DataTable();
* Wrapping table headers in a \<thead\> and the table body in a \<tbody\> ([required by Data Tables](https://datatables.net/manual/installation#HTML))
* Adding invisible table headers to the expandable/collapsible code context "table" (Brakeman::Report::HTML#with_context)

A nice thing added by this change is that, since we are making all the tables on the page sortable, any future tables that may get added are already made sortable too. If you add a table but don't wrap its headers in a \<thead\> and its body in a \<tbody\>, you will see the javascript error message

```
Datatables: Cannot read property 'mData' of undefined
```

So if you see this after adding a table, it means you forgot the wrapper tags.

I considered how to write tests for this, and thought it might be possible to write a Capybara+Poltergeist+PhantomJS test that does something like "When I click this table header, then I should see sorted data". What do you think, would that kind of test be useful? Please let me know and I will work on adding a few.

### DIFF
--- a/lib/brakeman/report/report_html.rb
+++ b/lib/brakeman/report/report_html.rb
@@ -139,6 +139,16 @@ class Brakeman::Report::HTML < Brakeman::Report::Base
     "<table id='#{code_id}' class='context' style='display:none'>" <<
     "<caption>#{CGI.escapeHTML warning_file(warning) || ''}</caption>"
 
+    output << <<-HTML
+      <thead style='display:none'>
+        <tr>
+          <th>line number</th>
+          <th>line content</th>
+        </tr>
+      </thead>
+      <tbody>
+    HTML
+
     unless context.empty?
       if warning.line - 1 == 1 or warning.line + 1 == 1
         error = " near_error"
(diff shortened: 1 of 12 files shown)


presidentbeef:
Hi David! Sorry I'm a little slow to respond. I'll check this out soon. The idea seems fine. I don't think tests are that necessary for this, since the HTML report isn't well tested anyway.

dlanner:
Cool, glad to hear it.

presidentbeef:
Again, my apologies for the delay. This looks great to me.

Thank you for your contribution!

dlanner:
No problem, glad to contribute!

Issue #723, opened 2015-09-21
Improve test coverage


presidentbeef:
Brakeman's test coverage is surprisingly not too bad, but it could be better! Some of the missing test cases are actually pretty straightforward to cover. Some suggestions are below.

**Improving Coverage**

[Current coverage via Code Climate](https://codeclimate.com/github/presidentbeef/brakeman/coverage?sort=covered_percent&sort_direction=asc).

Easy candidates:
-  ~~[lib/brakeman.rb](https://codeclimate.com/github/presidentbeef/brakeman/coverage/55ff8a70d86bab610f0320da)~~
- ~~[lib/brakeman/options.rb](https://codeclimate.com/github/presidentbeef/brakeman/coverage/55ff8a71d86bab610f03211d)~~

Medium candidates:
- [lib/brakeman/ignore/config.rb](https://codeclimate.com/github/presidentbeef/brakeman/coverage/55ff8a71d86bab610f03213f)
- [lib/brakeman/report/report_table.rb](https://codeclimate.com/github/presidentbeef/brakeman/coverage/55ff8a71d86bab610f032149) (Actually all the low coverage reports are the same issue)
- [lib/brakeman/util.rb](https://codeclimate.com/github/presidentbeef/brakeman/coverage/55ff8a71d86bab610f032154) (Several little missed bits)

**Running Tests**

After forking and cloning your fork of Brakeman, I recommend using rvm and creating a separate gemset for Brakeman with `rvm gemset use brakeman --create`.

Then:
- `gem install bundler` to get Bundler
- Run `bundle` to install dependencies
- Run `rake` to run tests
- Alternatively, use `ruby test/test.rb` to run tests
- Use `ruby test/test.rb --name test_name_here` to run a single test

**Adding/Modifying Tests**

Brakeman uses [test-unit](http://test-unit.github.io/#test-unit). Most of the tests are based around scanning applications in the [test/apps/](https://github.com/presidentbeef/brakeman/tree/master/test/apps) directory and then checking for expected warnings with `assert_warning` or `assert_no_warning`. The test infrastructure and coordination is in [test/test.rb](https://github.com/presidentbeef/brakeman/blob/master/test/test.rb).

The `BrakemanTester::RescanTestHelper` class can be useful for tweaking an existing application without changing the code in `test/apps`. See the [CVE tests](https://github.com/presidentbeef/brakeman/blob/master/test/tests/cves.rb) for an example of using rescanning to test specific versions.

[...]

PR #735, opened 2015-10-14
Add tests brakeman rb

bethanyr:
I have added a couple of tests for the brakeman.rb file to increase code coverage for issue #723. I plan to add more tests, but wanted to get feedback and make sure I am following the current pattern for testing.

### DIFF
--- a/test/tests/brakeman.rb
+++ b/test/tests/brakeman.rb
@@ -278,6 +278,11 @@ class ConfigTests < Test::Unit::TestCase
     output_format_tester({:output_files => ['xx.xx', 'xx.xx']}, [:to_s, :to_s])
     output_format_tester({:output_files => ['xx.html', 'xx.pdf', 'xx.csv', 'xx.tabs', 'xx.json', 'xx.md']}, [:to_html, :to_pdf, :to_csv, :to_tabs, :to_json, :to_markdown])
   end
+
+  def test_output_format_errors_raised
+    options = {:output_format => :to_json, :output_files => ['xx.csv', 'xx.xxx']}
+    assert_raises("ArgumentError: Cannot specify output format if multiple output files specified") {Brakeman.get_output_formats(options)}
+  end
 end
 
 class GemProcessorTests < Test::Unit::TestCase
--- a/test/tests/brakeman.rb
+++ b/test/tests/brakeman.rb
@@ -300,18 +300,33 @@ class ConfigTests < Test::Unit::TestCase
   end
 
   def test_github_options_returns_url
-    config = Tempfile.new("config")
+    # config = Tempfile.new("config")
 
-    config.write <<-YAML.strip
-    ---
-    :quiet: true
-    YAML
+    # config.write <<-YAML.strip
+    # ---
+    # :quiet: true
+    # YAML
 
-    config.close
+    # config.close
     options = {:github_repo => 'presidentbeef/brakeman', :app_path => "/tmp"}
 
     final_options = Brakeman.set_options(options)
     assert final_options[:github_url], "https://www.github.com/presidentbeef/brakeman"
+  end
+
+  def test_optional_check_options
+    options = {:list_optional_checks => true}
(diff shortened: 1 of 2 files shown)


presidentbeef:
Hi Bethany,

Thank you for your work on this. The style is fine. However, unless I'm missing something, you shouldn't need to create a config file for the tests you added. That was in the other tests because they were specifically testing behavior when reading config files.

Otherwise, great!

test/tests/brakeman.rb
```diff
@@ -278,6 +278,49 @@ def test_output_format
...
     output_format_tester({:output_files => ['xx.html', 'xx.pdf', 'xx.csv', 'xx.tabs', 'xx.json', 'xx.md']}, [:to_html, :to_pdf, :to_csv, :to_tabs, :to_json, :to_markdown])
   end
+
+  def test_output_format_errors_raised
+    options = {:output_format => :to_json, :output_files => ['xx.csv', 'xx.xxx']}
+    assert_raises("ArgumentError: Cannot specify output format if multiple output files specified") {Brakeman.get_output_formats(options)}
+  end
+
+  def test_github_options_raises_error
+    config = Tempfile.new("config")
```
presidentbeef: No need for the config file.

test/tests/brakeman.rb
```diff
@@ -278,6 +278,49 @@ def test_output_format
...
...t escape single quotes (CVE-2012-3464)\nCheckSkipBeforeFilter         Warn when skipping CSRF or authentication checks by default\nCheckSQL                      Check for SQL injection\nCheckSQLCVEs                  Checks for several SQL CVEs\nCheckSSLVerify                Checks for OpenSSL::SSL::VERIFY_NONE\nCheckStripTags                Report strip_tags vulnerabilities CVE-2011-2931 and CVE-2012-3465\nCheckSymbolDoSCVE             Checks for versions with ActiveRecord symbol denial of service vulnerability\nCheckTranslateBug             Report XSS vulnerability in translate helper\nCheckUnsafeReflection         Checks for unsafe reflection\nCheckValidationRegex          Report uses of validates_format_of with improper anchors\nCheckWithoutProtection        Check for mass assignment using without_protection\nCheckXMLDoS                   Checks for XML denial of service (CVE-2015-3227)\nCheckYAMLParsing              Checks for YAML parsing vulnerabilities (CVE-2013-0156)\nCheckSymbolDoS                Checks for symbol denial of service\nCheckUnscopedFind             Check for unscoped ActiveRecord queries\nCheckWeakHash                 Checks for use of weak hashes like MD5\n"
```
presidentbeef: I'm worried about maintaining this. The list of checks is only going to grow.

What about something like this (untested)?

```ruby
check_list = capture_output { Brakeman.list_checks(options) }
Brakeman::Checks.checks.each do |check|
  assert check_list.include? check.class.to_s
  assert check_list.include? check.description
end
```

Same with the optional checks.

bethanyr: good suggestion. I have updated the 2 tests to get the list of checks from Brakeman::Checks. There is currently a private method: get_check_name that is in Brakeman::Checks - which has the same code: to_s.split("::").last that I am using in the test. Do you have any objections if I make the method public to call it from the test? Or is the test code OK?

test/tests/brakeman.rb
bethanyr: good catch - I've removed it

test/tests/brakeman.rb
presidentbeef: You can make it public.

test/tests/brakeman.rb
```diff
@@ -278,6 +278,75 @@ def test_output_format
...
+      assert check_list.include? check.description
+    end
+  end
+
+  def test_dump_config_no_file
+    options = {:create_config => true, :test_option => "test"}
+    config_output = capture_output {
+      Brakeman.dump_config(options)
+    }[1]
+    assert_equal config_output, "---\n:test_option: test\n"
```
presidentbeef: It would be even cooler if this loaded the YAML and compared it to the original `options`

presidentbeef:
Hi Bethany, still working on this or ready to merge?

bethanyr:
Hi Justin - I think it's ready to merge. Currently lib/brakeman.rb is at 88% coverage.

presidentbeef:
Thanks! That put us up over 90%!

PR #741, opened 2015-10-16
Use SafeYAML to load configuration files

presidentbeef:


bf4:
If you don't own your configuration files you're in trouble safeyaml can't help you with :)  (Also, Ruby 2.0+ Has `YAML.safe_load`)

presidentbeef:
This is mostly for people providing Brakeman as a Service.

bf4:
So maybe just check if YAML responds to 'safe_load'?

B mobile phone

presidentbeef:
To what end? That doesn't help people using Ruby < 2.0. Not everyone is running Brakeman on trusted code. It is unexpected behavior for a static analysis tool to execute any of the code being scanned and I greatly prefer to avoid it.

PR #756, opened 2015-10-31
Make sure user_input is nil not false

presidentbeef:


presidentbeef:
This was caused by #728 and a lack of tests.

PR #758, opened 2015-11-02
Add Code Climate support

mrb:
This PR makes Brakeman compliant with the Code Climate engine specification, allowing it to be used as part of analysis on codeclimate.com and in the [CLI](http://github.com/codeclimate/codeclimate). As an overview, it:
- Adds a Dockerfile for packaging
- Adds a Code Climate reporter
- Adds a data file mapping remediation points to Brakeman issues
- Moves documentation from documentation site into the Brakeman repository


lib/brakeman/report/report_codeclimate.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def generate_report
+    all_warnings.map { |warning| issue_json(warning) }.join("\0")
+  end
+
+  private
+
+  def issue_json(warning)
+    check = warning.check.sub("Brakeman::Check", "")
+    path = warning.file.sub("#{tracker.app_path}/", "")
```
presidentbeef: Should be able to use `warning.relative_path` here.

test/tests/codeclimate_output.rb
```diff
@@ -0,0 +1,17 @@
...
+      :run_all_checks => true
+    ).report.to_codeclimate
+  end
+
+  def test_issue_output_is_parseable_as_json
+    issues = @@report.split("\0")
+
+    assert JSON.parse(issues[0]).is_a? Hash
+    assert JSON.parse(issues[1]).is_a? Hash
+    assert JSON.parse(issues[2]).is_a? Hash
```
presidentbeef: I suggest checking for presence of expeced keys, like in https://github.com/presidentbeef/brakeman/blob/master/test/tests/json_output.rb

docs/warning_types/cross-site_request_forgery/index.markdown
```diff
@@ -0,0 +1,10 @@
+Cross-site request forgery is #5 on the [OWASP Top Ten](https://www.owasp.org/index.php/Top_10_2010-A5). CSRF allows an attacker to perform actions on a website as if they are an authenticated user.
+
+This warning is raised when no call to `protect_from_forgery` is found in `ApplicationController`. This method prevents CSRF.
+
+For Rails 4 applications, it is recommended that you use `protect_from_forgery :with => :exception`. This code is inserted into newly generated applications. The default is to `nil` out the session object, which has been a source of many CSRF bypasses due to session memoization.
+
+See [the Ruby Security Guide](http://guides.rubyonrails.org/security.html#cross-site-request-forgery-csrf) for details.
+
+---
+Back to [Warning Types](/docs/warning_types)
```
presidentbeef: Missed link

docs/warning_types/command_injection/index.markdown
```diff
@@ -0,0 +1,16 @@
...
+    system("ls #{params[:dir]}")
+
+    exec("md5sum #{params[:input]}")
+
+Brakeman will warn on any method like these that uses user input or unsafely interpolates variables.
+
+See [the Ruby Security Guide](http://guides.rubyonrails.org/security.html#command-line-injection) for details.
+
+---
+Back to [Warning Types](/docs/warning_types)
```
presidentbeef: Missed link

docs/warning_types/cross_site_request_forgery/index.markdown
```diff
@@ -0,0 +1 @@
+../cross-site_request_forgery/index.markdown
```
presidentbeef: I believe the files with content should match the actual category names.

docs/warning_types/mass_assignment/index.markdown
```diff
@@ -0,0 +1,31 @@
...
+
+    config.active_record.whitelist_attributes = true
+
+Unfortunately, it can also easily be bypassed:
+
+    User.new(params[:user], :without_protection => true)
+
+Brakeman will warn on uses of `without_protection`.
+
+Please see [this page](http://railspikes.com/2008/9/22/is-your-rails-application-safe-from-mass-assignment) for more details.
```
presidentbeef: This link was recently removed from the website because railspikes.com does not appear to exist any longer.

lib/brakeman/report/report_codeclimate.rb
```diff
@@ -0,0 +1,70 @@
...
+    body = content_for(warning.warning_type, warning.link)
+
+    {
+      type: "Issue",
+      check_name: check,
```
presidentbeef: Is this a value that's displayed to end users? Or used to control something programmatically (like turning off checks)?

mrb: @presidentbeef It's used to turn of checks, and is only incidentally displayed to users. We're changing it to map to the warning code name to be consistent.

mrb:
@presidentbeef Thanks for your comments - will be back with an update soon.

lib/brakeman/report/report_codeclimate.rb
presidentbeef: For turning off checks you'll want the name of the class like you have it already.

presidentbeef: Unless I'm misunderstanding how you will map it. Otherwise I'd recommend going with the `warning_type` which is more like a category.

mrb:
@presidentbeef Okay! That should cover your comments. Let me know if there's anything else we need to address.

test/tests/codeclimate_output.rb
```diff
@@ -0,0 +1,22 @@
+class TestCodeClimateOutput < Test::Unit::TestCase
+  def setup
+    @@report ||= Brakeman.run("#{TEST_PATH}/apps/rails2").report.to_codeclimate
+    @@issue ||= JSON.parse(@@report.split("\0").first)
```
presidentbeef: These tests should check every issue, not just the first one.

lib/brakeman/report/report_codeclimate.rb
jpignata: @presidentbeef We use the `check_name` value to suppress issues in our UI. It's surfaced to users in that they must configure the `check_name` values that they'd like to suppress in their .codeclimate.yml file.

Suppressed checks are still run, we just don't surface the resulting issues to a user. Our thought for using the name for a `warning_code` here instead of the class name of a check is that since a check can emit one of several warning codes, it's safer to have a user intentional ignore each `warning_code` rather than an entire check. Does that make sense?

test/tests/codeclimate_output.rb
jpignata: @presidentbeef Right on. Fixed.

lib/brakeman/report/report_codeclimate.rb
presidentbeef: Yes, that makes sense. If the filtering is done on your side, `warning_code` is the best.

jpignata: Awesome. I've made that change. Thanks for all of the feedback. Let me know if there's anything else pending or other changes you'd like to see.

lib/brakeman.rb
```diff
@@ -180,6 +180,8 @@ def self.get_formats_from_output_format output_format
       [:to_json]
     when :markdown, :to_markdown
       [:to_markdown]
+    when :cc, :codeclimate, :to_codeclimate
```
presidentbeef: Having `:cc` here doesn't do anything unless it's added to the options parser as well.

presidentbeef: Do you want people to be able to do `brakeman -o report.cc`? If so it should be added below in `get_formats_from_output_files`.

presidentbeef:
Have you done any larger scale testing? Is there a validation tool for the platform spec?

lib/brakeman.rb
jpignata: Yes, I think that would be helpful. I've added the branch to `get_formats_from_output_files`. I've also fixed the `:cc` option – thanks!

presidentbeef:
Thanks!

presidentbeef:
@mrb / @ABaldwinHunter does this match CC behavior better?

PR #759, opened 2015-11-03
Remove stray report output

jpignata:
We left a stray report file in the project in our branch during testing and it was merged before we caught it.


presidentbeef:
Oops.

PR #764, opened 2015-11-10
Add support for directories to the only-files and skip-files options

ptoomey3:
This is a quick and dirty spike for adding first-class directory support to the `only-files` and `skip-files` options. The prior logic vaguely supported directories for the `only-files` option, but not with the `skip-files` option (since that option forced a `$` regex anchor on the match). I debated on whether it made sense to overload the `only-files`/`skip-files` options or if this was different enough to add a `only-directories`/`skip-directories` option. I decided that a new option felt too heavy, but I welcome feedback. The only other quirky bit to this is what I expect to be passed to these options to convey that they are in fact directories. My current implementation does the following:

* Force all directory matches to have a trailing file path separator (ex `/`). If we don't do this then it is tricky to distinguish between a file match and a directory path match.
* Allow a leading file path separator to tell us that the match should only be made from the application relative root directory (i.e. is a prefix match). This is useful when you want to exclude directories, but don't want to go overboard. For example, I might want to exclude the "test" directory from the root of my project, but not exclude the "test" directory in "lib/test/some_test_lib_thing.rb". In short, if you don't want your path to match an arbitrary part of a path then you can prefix it with a file path separator to force the directory to be a prefix match (ex. `/test/`). 
* I am now only matching on directories within the project's directory space. Since we are using `/test` to mean "the test directory directly off of my application's root" it makes sense to only match on the project's directory space. The prior logic would have allowed something like `/Users/ptoomey3/my_rails_project/test` to do the equivalent. That would no longer work, and could be backward incompatible. But, I imagine that could be rare, since that kind of hard coded path is relatively fragile, and I imagine most folks were using something like `/test` to begin with. 

@presidentbeef - If this seems like something you would consider merging just let me know if the current direction is ok. And, if you have any recommendations on the best way to add tests for such an option that would be great (ex. do you think it would be best to just add on more files/paths to the existing `test/tests/only_file_options.rb` test?). 

/cc @mastahyeti @oreoshake


lib/brakeman/app_tree.rb
```diff
@@ -10,10 +12,47 @@ def self.from_options(options)
...
+            "#{Regexp.escape f}\\z"
+          end
+        end
+        init_options[:skip_files] = Regexp.new("(?:" << skip_files_escaped.join("|") << ")")
       end
+
       if options[:only_files]
-        init_options[:only_files] = Regexp.new("(?:" << options[:only_files].map { |f| Regexp.escape f }.join("|") << ")")
+        only_files_escaped = options[:only_files].map do |f|
+          # If it ends in a file separator then we assume it is a path rather
```
ptoomey3: This logic is duplicated once for each option. It could probably be folded down into a single method `self.regexes_for_paths` or some such thing.

presidentbeef:
I haven't read the code carefully but yes I think the functionality makes sense and it's fine for these options to work this way.

presidentbeef:
@mrb / @ABaldwinHunter does this match CC behavior better?

brakeman.gemspec
```diff
@@ -15,6 +15,7 @@ Gem::Specification.new do |s|
   s.cert_chain  = ['brakeman-public_cert.pem']
   s.signing_key = gem_priv_key if File.exist? gem_priv_key and $0 =~ /gem\z/
   s.add_development_dependency "test-unit"
+  s.add_development_dependency "byebug"
```
ptoomey3: I'm obviously fine with ripping this out, but I always find myself adding this line temporarily when working on brakeman. I'm not at all committed to byebug specifically, but bundling in a reasonable debugger is nice for dev.

test/tests/only_files_option.rb
```diff
@@ -19,7 +19,7 @@ def expected
   end
 
   def report
-    @@report ||= BrakemanTester.run_scan "rails3.2", "Rails 3.2", { :only_files => ["app/views/users/"], :skip_files => ["app/views/users/sanitized.html.erb"] }
+    @@report ||= BrakemanTester.run_scan "rails3.2", "Rails 3.2", { :only_files => ["app/views/users/", "exec_controller.rb", "/app/models/user/"], :skip_files => ["app/views/users/sanitized.html.erb"] }
```
ptoomey3: While this doesn't exercise 100% of the new logic, this seems to provide a good sense of things working as expected without dramatically refactoring this test suite. If you would like to see something else added just let me know.

ptoomey3:
Bah, I forgot that byebug only runs on ruby 2.X. I'll rip it out for now. I think this is looking pretty good and ready for :eyes:.

lib/brakeman/app_tree.rb
```diff
@@ -96,13 +128,28 @@ def select_files(paths)
...
     def select_only_files(paths)
       return paths unless @only_files
-      paths.select { |f| @only_files.match f }
+      project_root  = Pathname.new(@root)
+      paths.select do |path|
+        absolute_path = Pathname.new(path)
+        # relative root never has a leading separator. But, we use a leading
+        # separator in a @skip_files entry to imply that a directory is
+        # "absolute" with respect to the project directory.
+        project_relative_path = File::SEPARATOR + absolute_path.relative_path_from(project_root).to_s
```
presidentbeef: I think `File.join` is better here.

lib/brakeman/app_tree.rb
```diff
@@ -96,13 +128,28 @@ def select_files(paths)
...
     def reject_skipped_files(paths)
       return paths unless @skip_files
-      paths.reject { |f| @skip_files.match f }
+      project_root  = Pathname.new(@root)
+      paths.reject do |path|
+        absolute_path = Pathname.new(path)
+        # relative root never has a leading separator. But, we use a leading
+        # separator in a @skip_files entry to imply that a directory is
+        # "absolute" with respect to the project directory.
+        project_relative_path = File::SEPARATOR + absolute_path.relative_path_from(project_root).to_s
```
presidentbeef: `File.join` here, too.

lib/brakeman/app_tree.rb
ptoomey3: Totally..I think my mind lapsed and forgot about `File.join` all together.

presidentbeef:
@ptoomey3 can you make those `File.join` changes? Thanks!

ptoomey3:
Doh...sorry..totally fell of my radar. Thanks for the bump. I'll do this as soon as I get back next week.

lib/brakeman/app_tree.rb
ptoomey3: Hmm...ok...maybe I wasn't totally crazy here :smile:. Reading my own comment above...we are forcing a leading `File::SEPERATOR` prefix since that is what we use in https://github.com/presidentbeef/brakeman/pull/764/files#diff-d300ce9778018ef3e26c917b83c13230R38 to delineate a "project relative root". And, by default, `File.join` doesn't return a leading slash. So, in order to get `File.join` to do that you have to call it using a somewhat ugly/non-obvious approach:

``` ruby
File.join("foo/bar")
=> "foo/bar"
File.join("", "foo/bar")
=> "/foo/bar"
File.join(File::SEPARATOR, "foo/bar")
=> "/foo/bar"
```

All of the approaches to force the leading slash seem pretty non-obvious (i.e. it seems to be abusing what `File.join` actually is supposed to be doing. Thoughts?

presidentbeef: I'd be okay with either of the last two options.

ptoomey3: Updated in https://github.com/presidentbeef/brakeman/commit/69057b386b7d717a21a1c50902943b927c27e078.

presidentbeef:
Thanks, Patrick! Sorry for the delay.

ptoomey3:
No worries. Thanks for brakeman :smile:

Issue #770, opened 2015-11-25
Brakeman invoked from command line does not load config file


carmi:
I was investigating some weirdness where when running brakeman from the command line with `bundle exec brakeman -c myconfig.yml` it would respect the settings in myconfig.yml for all options except the `exit_on_warn` option.

Further investigation revealed that the loading of the `options[:config_file]` occurs in Brakeman.run and never propogates back to the bin/brakeman file. So on [line 81 of bin/brakeman](https://github.com/presidentbeef/brakeman/blob/master/bin/brakeman#L81) where it checks if `options[:exit_on_warn]` is present, it only respects this flag being passed in the command line options and not being set in the config file.

This seems like a bug to me - I think if you define exit_on_warn in your config file, then running brakeman from the command line should respect that and exit an error status code if warnings are found.

Before I go off and create a pull request, I wanted to check if this is expected behavior for some reason that I am not thinking of, or if in fact this is a bug?

** Repetition Steps **

With a rails app with warnings, and a config `brakeman.yml` with the following:

```
---
:exit_on_warn: true
:min_confidence: 2
```

Running 

`brakeman -c brakeman.yml; echo $?` will output exit code 0.

But running

`brakeman -c brakeman.yml -z; echo $?` will output exit code 3.

I think these should both return exit code 3.

carmi:
I threw together a rough fix at https://github.com/presidentbeef/brakeman/pull/771 to help demonstrate issue.

presidentbeef:
Yes, definitely a bug. Thanks for reporting!

PR #771, opened 2015-11-25
Load options including config_file options

carmi:
An example fix for Issue: https://github.com/presidentbeef/brakeman/issues/770

Previously, the config_file options were not being loaded inside the
bin/brakeman context.

This is probably not as clean as it could be - but demonstrates a fix to the issue.


presidentbeef:
Thanks, Evan. I went with #774 to resolve the issue.

PR #780, opened 2015-12-04
Address issue #723, increase lib/options.rb test coverage

Zanadar:
I've added a series of tests which check for population of the options hash based in command line input and flags. This PR aims to address #723. I'd love your feedback.  Thanks!

### DIFF
--- a/test/tests/options.rb
+++ b/test/tests/options.rb
@@ -1,0 +1,291 @@
+require 'brakeman/options'
+
+class BrakemanOptionsTest < Test::Unit::TestCase
+  EASY_OPTION_INPUTS = {
+    :exit_on_warn           => "-z",
+    :rails3                 => "-3",
+    :run_all_checks         => "-A",
+    :assume_all_routes      => "-a",
+    :escape_html            => "-e",
+    :ignore_model_output   => "--ignore-model-output",
+    :ignore_attr_protected  => "--ignore-protected",
+    :interprocedural        => "--interprocedural",
+    :ignore_ifs             => "--no-branching",
+    :skip_libs              => "--skip-libs",
+    :debug                  => "-d",
+    :interactive_ignore     => "-I",
+    :report_routes          => "-m",
+    :summary_only           => "--summary",
+    :absolute_paths         => "--absolute-paths",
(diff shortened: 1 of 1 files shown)


test/tests/options.rb
```diff
@@ -0,0 +1,291 @@
...
+    options = setup_options_from_input("-4")
+    assert options[:rails4] && options[:rails3]
+
+    options = setup_options_from_input("--rails4")
+    assert options[:rails4] && options[:rails3]
+  end
+
+  def test_safe_methods_option
+    options = setup_options_from_input("--safe-methods", "test_method2,test_method1,test_method2")
+    assert options[:safe_methods] == Set.new([:test_method1, :test_method2])
```
presidentbeef: When doing comparisons, `assert_equal` would be the typical approach (and produces better failure messages). Expected value comes first.

And you can use `Set[]` to create the set.

```ruby
assert_equal Set[:test_method1, :test_method2], options[:safe_methods]
```

test/tests/options.rb
```diff
@@ -0,0 +1,291 @@
...
+    format_options.each_pair do |key, value|
+      options = setup_options_from_input("--format", "#{key}")
+      assert options[:output_format] == value, "Expected #{key} to be #{value}."
+    end
+  end
+
+  def test_CSS_file_option
+    options = setup_options_from_input("--css-file", "../test.css")
+    local_path = File.expand_path('../test.css')
+    assert options[:html_style], local_path
```
presidentbeef: Did you mean to use `assert_equal` here (and in a bunch of other places below)? The second argument to `assert` is the failure message.

presidentbeef:
Hi Zander, other than some issues with `assert` usage, this looks really nice!

It seems only `-r`, `-o`, and the combination of `--compare file.json -o diff.json` are missing.

test/tests/options.rb
Zanadar: Thanks for this note (and below). Will push an update.

Zanadar:
Thanks much for the feedback (and good catch on the missing options...).  Will push an update asap.

presidentbeef:
Nice, thanks! :zap:

PR #781, opened 2015-12-04
Ignore secrets.yml not in source control

### LINKED ISSUE #777: False positive on secrets.yml check
The fix for #760 introduces a new false positive.

On my project, the secrets.yml file is not checked into version control (and is ignored in .gitignore), but the security warning is raised.

presidentbeef:
fixes #777


jgaxn:
This is definitely better. However, in my projects I use `/config/*.yml` in my .gitignore and I still get the false postitive. Maybe something in `ignored?` to look for matching wildcards?

presidentbeef:
Eh, I'll just add `config/*.yml` directly.

jgaxn:
Awesome, thanks!

PR #784, opened 2015-12-07
Fix various Ruby warnings

andyw8:
I saw your tweet! https://twitter.com/brakeman/status/672940088175628288

### DIFF
--- a/lib/brakeman/checks/check_cross_site_scripting.rb
+++ b/lib/brakeman/checks/check_cross_site_scripting.rb
@@ -105,12 +105,6 @@ class Brakeman::CheckCrossSiteScripting < Brakeman::BaseCheck
           warning_code = :xss_to_json
         end
 
-        code = if match == out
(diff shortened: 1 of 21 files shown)


lib/brakeman/checks/check_cross_site_scripting.rb
```diff
@@ -115,7 +115,7 @@ def check_for_immediate_xss exp
           :warning_type => "Cross Site Scripting",
           :warning_code => warning_code,
           :message => message,
-          :code => match,
+          :code => code,
```
andyw8: You'll probably want to verify my assumption here. If `:code` *is* supposed to be `match` then the `code` assignment just above should be removed.

andyw8: Oops, ignore my previous comment.

presidentbeef:
Thanks!

PR #788, opened 2015-12-16
Emit brakeman's native fingerprints for Code Climate engine.

noahd1:
Code Climate has fallback fingerprints for issues emitted by engines,
but it's much preferable to use the fingerprints that brakeman
itself creates. The fingerprints take into account a lot more
criteria, are maintained and may be already used by customers
in their brakeman.ignore file.

/c @codeclimate/review


noahd1:
@presidentbeef ping

presidentbeef:
I don't have much of an opinion here :)

noahd1:
Figured! Thanks.

PR #802, opened 2016-01-27
Fix CodeClimate include_paths behavior

wfleming:
The chopping we did on these paths was leading to unexpected results against master. Probably related to changes in 656db4bbdb7fd7: since we always add `/` to the end of directories when we construct `include_paths` before we run the engine, we can now pass through `include_paths` from `/config.json` effectively unmodified.

### DIFF
--- a/bin/codeclimate-brakeman
+++ b/bin/codeclimate-brakeman
@@ -12,10 +12,6 @@ end
 command = "/usr/src/app/bin/brakeman --quiet --format codeclimate"
 
 if engine_config["include_paths"]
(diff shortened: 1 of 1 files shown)


presidentbeef:
Gah that's why I tried to ping you guys on #764.

wfleming:
@presidentbeef Oops! Sorry, our bad: the notifications must have been missed over here.

presidentbeef:
It's alright, both changes go out in the next release so shouldn't have impacted anyone yet.

PR #810, opened 2016-01-28
Replaced some :!= with :"!=" to fix "syntax error, unexpected tNEQ" in Ruby 1.8

roniegh:


presidentbeef:
Hi, sorry, but Ruby 1.8 is no longer supported as of Brakeman 3.1.0: http://brakemanscanner.org/blog/2015/08/30/brakeman-3-dot-1-0-released/

PR #811, opened 2016-01-28
Replaced key: "value" with :key => "value" hash syntax to fix some Ruby 1.8 syntax errors

roniegh:


presidentbeef:
Hi, sorry, but Ruby 1.8 is no longer supported as of Brakeman 3.1.0: http://brakemanscanner.org/blog/2015/08/30/brakeman-3-dot-1-0-released/

roniegh:
@presidentbeef oops! Sorry, my bad. Thanks for the quick reply.

By the way, brakeman-min 3.1.5 still works fine on Ruby 1.8 after applying 44eef2ad044e9348fb4450ca7819e47f2cafe2b7 + 70630732a3e4bfec80f8d82ad691544346a0e66e also passed all tests on Ruby 1.9.3, 2.2.2, 2.3.

presidentbeef:
That's cool. I just don't want to have to limit myself or any contributions to 1.8 syntax and features any longer.

Given Ruby 1.8 and 1.9 are both essentially dead and unsupported, I hope you can upgrade soon. There is no reason to run Brakeman with 1.8 unless for some reason you can't run a newer Ruby.

PR #815, opened 2016-02-04
Index calls in class bodies

### LINKED ISSUE #814: Have brakeman index function calls in class definitions
Brakeman doesn't index things outside of methods but inside of classes. 

```ruby

class SomethingChill
  SomethingNotSoChill.exec()  # no warning 

  def stuff
    SomethingElseNotSoChill.exec() # :boom:
  end
end
```

In this case, brakeman finds the call inside of the definition but does not find the call in the class def.

presidentbeef:
fixes #814 

/ @oreoshake


oreoshake:
:fireworks: :confetti_ball:

PR #830, opened 2016-02-25
MultiJson is not a brakeman dependency

bf4:
Using it in the `bin/brakeman` causes installs outside of the Gemfile to also install `multi_json`

It is a transient dependency of the test apps via JBuilder via Rails.

This is causing failures in https://travis-ci.org/rubygems/rubygems.org/builds/111655610

ref: https://github.com/rubygems/rubygems.org/pull/1201

### DIFF
--- a/bin/brakeman
+++ b/bin/brakeman
@@ -2,6 +2,7 @@
 #Adjust path in case called directly and not through gem
 $:.unshift "#{File.expand_path(File.dirname(__FILE__))}/../lib"
 
+require 'json'
 require 'brakeman'
 require 'brakeman/options'
 require 'brakeman/version'
@@ -62,12 +63,12 @@ begin
 
     if options[:comparison_output_file]
       File.open options[:comparison_output_file], "w" do |f|
-        f.puts MultiJson.dump(vulns, :pretty => true)
+        f.puts Json.dump(vulns, :pretty => true)
       end
 
       Brakeman.notify "Comparison saved in '#{options[:comparison_output_file]}'"
     else
-      puts MultiJson.dump(vulns, :pretty => true)
+      puts Json.dump(vulns, :pretty => true)
     end
 
     if options[:exit_on_warn] && vulns[:new].count > 0


bf4:
Also, looks like brakeman-min should also depend on terminal-table and highline?

presidentbeef:
I guess I didn't test this, and then you didn't test your patch, and then I didn't test your patch before merging. :unamused:

presidentbeef:
`brakeman-min` has minimal dependencies so you can only install what you need. terminal-table and highline are not required for minimal operation.

bf4:
Wow, I'm pretty embarrassed that I mispeled jzon there. 

I can help write a test around the `bin/brakeman` usage.  BTW, if you're interested, `s/bin/exe` is the new orange is the new black http://bundler.io/blog/2015/03/20/moving-bins-to-exe.html

I'm happy to make that as a PR, as well.

But, down to business: What command-line usage is intended with `brakeman-min`?  Or is it intended to run via rake task? I had thought of it as brakeman without all the template rendering. Is the idea JSON only?

presidentbeef:
The purpose of `brakeman-min` is to allow people to only install the dependencies they need. As part of that, it outputs JSON by default (to avoid the `terminal-table` dependency), does not include any templating engines (since people often just use one), and does not have `highline` (since most people don't use `-I`). However, I think I'm going to avoid all this and just vendor all dependencies in the next release. Then no one will bug me about conflicting or out of date dependencies.

I see no reason to use `exe/` given Brakeman's gemspecs specify the executable, and I'm the only one who builds the gem and I don't use `bundler` in development, so I wouldn't have the problems that post describes.

Yes, tests for any `bin/brakeman` behavior that does not follow the normal path would be appreciated.

bf4:
Sorry, didn't mean to feel like I'm bugging you. It's a great lib and you
know it's one of the one's people depend on.

presidentbeef:
Didn't mean you in particular :) It's just an ongoing issue when people mix their application dependencies with things that aren't code dependencies - like Brakeman.

PR #843, opened 2016-03-22
Add `--no-output-files` option

GordonDiggs:
Permits a user to override output options in Brakeman config from command line, and run Brakeman without writing files to system.

This change lets Code Climate run with a line-item ignore of that part of a user's Brakeman config.

Authored with @gordondiggs

### DIFF
--- a/bin/codeclimate-brakeman
+++ b/bin/codeclimate-brakeman
@@ -10,7 +10,7 @@ else
   engine_config = {}
 end
 
-command = "/usr/src/app/bin/brakeman --quiet --format codeclimate"
+command = "/usr/src/app/bin/brakeman --quiet --no-output-files --format codeclimate"
 
 if engine_config["include_paths"]
   command += " --only-files " + engine_config["include_paths"].compact.join(",").shellescape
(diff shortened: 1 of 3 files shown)


presidentbeef:
Not really a fan...can't you just `-o /dev/null`?

GordonDiggs:
@presidentbeef That squashes the output of issues to stdout (which is where we read the results from), right? We can't do `-o /dev/stdout` because there is a bug in some Docker versions that gets in the way.

presidentbeef:
Ugh.

GordonDiggs:
@presidentbeef one other solution I've thought of, which is a bit messier but might work: we could parse the `./config/brakeman.yml` file and remove the output files from it if they're specified. Would that be better than this? It would definitely be localized to our wrapper rather than polluting the options

presidentbeef:
Something I was going to suggest previously is to change from invoking the command line app to just doing it all programmatically. In this case that means you could use methods in `lib/brakeman.rb` to pull in the configuration file, remove/modify as necessary, and then pass those options in directly.

GordonDiggs:
@presidentbeef yep that sounds like a better idea. Going to go that route and will open another PR soon. Thanks for talking through it!

Issue #848, opened 2016-03-31
CheckForgerySetting Skips Controllers


jsyeo:
Currently brakeman only performs `protect_from_forgery` checks in the `ApplicationController`. While this is sufficient for most cases, it fails to catch controllers that also inherit from `ActionController::Base` but use a different name. Like for example `BaseController` or `AdminController`. See https://github.com/typus/typus/blob/master/app/controllers/admin/base_controller.rb#L1 and https://github.com/spree/spree/blob/148bcaf1e44081f940248d94d3aa6c5e95bb4fe2/api/app/controllers/spree/api/base_controller.rb#L5.

This [commit](https://github.com/typus/typus/commit/8791933a08b5fca801898649fe7aa15e99436f1f) was recently added to [typus](https://github.com/typus/typus) to turn on CSRF protection. This shows that brakeman shouldn't just simply restrict itself to `ApplicationController`s.

presidentbeef:
Agree, related to #664.

PR #857, opened 2016-04-04
Add support for configurable engines path

### LINKED ISSUE #664: Brakeman does not catch lack of protect_from_forgery in Rails engines
It looks like the CheckForgerySetting will only check ApplicationController[1]. This becomes an issue when running the test against a Rails engine as the rails engine generator will throw the ApplicationController into a namespace (i.e. Foobar::ApplicationController) which the check will be unable to find. With help from @jjarmoc, we came up with putting the entire check into a block that grabs all ApplicationControllers via a regex:

``` ruby
  def run_check
    tracker.controllers.keys.select{|key| key =~ /.*ApplicationController/ }.each do |controller_key|
      app_controller = tracker.controllers[controller_key]

      return unless ancestor? app_controller, :"ActionController::Base"
      ...
```

This should catch CSRF when running brakeman against the Rails engine itself, however it becomes more complicated when attempting to find it in Rails engines that are in other Rails applications. It looks like engines used to be generated in the engines/ directory, but when following the documentation[2], the engine just shows up in the root as a directory with whatever name you give it. A possible way to detect this is via the routes file where AFAIK is how you tell Rails you want to use the engine using something like `mount Blorgh::Engine => "/blorgh"`. This should tell you the Rails engines being used but I'm not sure how you would differentiate between a Rails engine that's a part of the source and a Rails engine that's coming from a gem. Maybe checking for the `blorgh` directory in the root and if it's not there assume it's coming from a gem?

[1] https://github.com/presidentbeef/brakeman/blob/a3dab5526275d86188b77a61883d69353291aa1e/lib/brakeman/checks/check_forgery_setting.rb#L13
[2] http://guides.rubyonrails.org/engines.html#generating-an-engine

jsyeo:
This PR is to add support for scanning engines that are not on the `engines/` path.

This is the first step needed to fix #664. See https://github.com/presidentbeef/brakeman/issues/664#issuecomment-112504060.

> The next step would probably be adding an option so you could point it at the engine directory - that would allow for pointing it at gems too.

This PR adds an `engines_path` option to brakeman to allow users to add engines that the application uses. The path to the engines can either be absolute or relative to the project root. For example, if the `brakeman.yml` config file contains this:

```
:engines_path:
  - engines/user_removal
```

Brakeman will scan the `engines/user_removal` directory relative to the project root. If the path starts with a forward slash, we assume that the path is relative. Like so:

```
:engines_path:
  - /path/to/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/administrate-0.1.4
```

### DIFF
--- a/lib/brakeman/app_tree.rb
+++ b/lib/brakeman/app_tree.rb
@@ -19,6 +19,7 @@ module Brakeman
         init_options[:only_files] = regex_for_paths(options[:only_files])
       end
       init_options[:additional_libs_path] = options[:additional_libs_path]
+      init_options[:engines_path] = options[:engines_path]
       new(root, init_options)
     end
 
@@ -56,6 +57,9 @@ module Brakeman
       @skip_files = init_options[:skip_files]
       @only_files = init_options[:only_files]
       @additional_libs_path = init_options[:additional_libs_path] || []
+      @engines_path = init_options[:engines_path] || []
+      @absolute_engines_path = @engines_path.select { |path| path.start_with?(File::SEPARATOR) }
+      @relative_engines_path = @engines_path - @absolute_engines_path
     end
 
     def expand_path(path)
@@ -100,8 +104,7 @@ module Brakeman
     end
 
     def layout_exists?(name)
-      pattern = "#{@root}/{engines/*/,}app/views/layouts/#{name}.html.{erb,haml,slim}"
-      !Dir.glob(pattern).empty?
+      !find_file("app/views/layouts", name, "{.erb, .haml, .slim}").empty?
     end
 
     def lib_paths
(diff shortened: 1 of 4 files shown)


jsyeo:
> This is the first step needed to fix #664.

I guess the next step needed would be to fix #848.

presidentbeef:
Needs to be added as an actual commandline option :) in https://github.com/presidentbeef/brakeman/blob/master/lib/brakeman/options.rb

jsyeo:
@presidentbeef added. :)

presidentbeef:
Hi Jason, sorry for the delay on this. Committing to reviewing your pull requests tomorrow (Friday).

jsyeo:
Thanks! No worries. :wink:

presidentbeef:
The `engines` directory should be included by default. For testing, I would add a non-standard engines directory to `rails4_with_engines`.

Additionally, the option should be able to specify the top of the directory (e.g. `engines` not just `engines/some_engine`). I think this would already work, but just pointing it out since your test case does not work that way.

presidentbeef:
Merging with #948

PR #858, opened 2016-04-04
Check forgery setting in all direct subclasses of ActionController::Base

### LINKED ISSUE #664: Brakeman does not catch lack of protect_from_forgery in Rails engines

### LINKED ISSUE #848: CheckForgerySetting Skips Controllers

jsyeo:
This fixes #848. This is the second part to fix #664. Instead of just scanning controllers named `ApplicationController`, we scan direct subclasses of `ActionController::Base` instead.

Once we have both this PR and #857 merged, we can close #664. :wink:

### DIFF
--- a/lib/brakeman/checks/check_forgery_setting.rb
+++ b/lib/brakeman/checks/check_forgery_setting.rb
@@ -7,67 +7,69 @@ require 'brakeman/checks/base_check'
 class Brakeman::CheckForgerySetting < Brakeman::BaseCheck
   Brakeman::Checks.add self
 
-  @description = "Verifies that protect_from_forgery is enabled in ApplicationController"
+  @description = "Verifies that protect_from_forgery is enabled in direct subclasses of ActionController::Base"
 
   def run_check
-    app_controller = tracker.controllers[:ApplicationController]
-    return unless app_controller and app_controller.ancestor? :"ActionController::Base"
+    tracker.controllers
+        .select { |_, controller| controller.parent == :"ActionController::Base" }
+        .each do |name, controller|
 
-    if tracker.config.allow_forgery_protection?
-      warn :controller => :ApplicationController,
-        :warning_type => "Cross-Site Request Forgery",
-        :warning_code => :csrf_protection_disabled,
-        :message => "Forgery protection is disabled",
-        :confidence => CONFIDENCE[:high],
-        :file => app_controller.file
+      if tracker.config.allow_forgery_protection?
+        warn :controller => name,
+             :warning_type => "Cross-Site Request Forgery",
+             :warning_code => :csrf_protection_disabled,
+             :message => "Forgery protection is disabled",
+             :confidence => CONFIDENCE[:high],
+             :file => controller.file
 
-    elsif app_controller and not app_controller.protect_from_forgery?
+      elsif controller and not controller.protect_from_forgery?
 
-      warn :controller => :ApplicationController,
-        :warning_type => "Cross-Site Request Forgery",
-        :warning_code => :csrf_protection_missing,
-        :message => "'protect_from_forgery' should be called in ApplicationController",
-        :confidence => CONFIDENCE[:high],
-        :file => app_controller.file,
-        :line => app_controller.top_line
+        warn :controller => name,
+             :warning_type => "Cross-Site Request Forgery",
+             :warning_code => :csrf_protection_missing,
+             :message => "'protect_from_forgery' should be called in #{name}",
+             :confidence => CONFIDENCE[:high],
+             :file => controller.file,
+             :line => controller.top_line
 
-    elsif version_between? "2.1.0", "2.3.10"
+      elsif version_between? "2.1.0", "2.3.10"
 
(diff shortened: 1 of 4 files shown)


lib/brakeman/checks/check_forgery_setting.rb
```diff
@@ -7,67 +7,69 @@
...
 
-    if tracker.config.allow_forgery_protection?
-      warn :controller => :ApplicationController,
-        :warning_type => "Cross-Site Request Forgery",
-        :warning_code => :csrf_protection_disabled,
-        :message => "Forgery protection is disabled",
-        :confidence => CONFIDENCE[:high],
-        :file => app_controller.file
+      if tracker.config.allow_forgery_protection?
+        warn :controller => name,
```
presidentbeef: I don't think we want to generate this warning for every controller.

jsyeo: It will only generate a warning for a controller that directly inherits from `ActionController::Base`. Will that be an issue? I would imagine that there may be apps that has more than one controllers that inherits from `ActionController::Base`. Like say the `ApplicationController` and `AdminController`.

presidentbeef: It's warning about a global setting (*that probably no one ever uses*), so it doesn't make sense to warn about it for every controller that inherits from `ActionController::Base`. Actually it doesn't make sense to tie it to a controller at all, but this code was written over five years ago and I didn't know what I was doing.

test/tests/rails4_with_engines.rb
```diff
@@ -6,7 +6,7 @@ class Rails4WithEnginesTests < Test::Unit::TestCase
 
   def expected
     @expected ||= {
-      :controller => 1,
+      :controller => 2,
```
presidentbeef: Please add a real test. Otherwise it's very hard to tell what broke. [Documentation here](https://github.com/presidentbeef/brakeman/wiki/How-to-Report-a-Brakeman-Issue#preparing-tests).

jsyeo: Added. https://github.com/presidentbeef/brakeman/pull/858/commits/60fdd48538353cf5176f2c7dd3f169d09d10455f

lib/brakeman/checks/check_forgery_setting.rb
jsyeo: I factored out that check from the `.each` iteration. One side effect of that is that for the `rails3.rb` test, we prioritize the warning of CVE20110447 over the missing `protect_from_forgery` call. See https://github.com/presidentbeef/brakeman/pull/858/files#diff-29d5ade8f0be60f47a3073ebab75b8d4L386. Which kind of makes sense in my opinion because there's no point asking the user to add the call to when rails need upgrading or patching in the first place.

presidentbeef:
Closing, will merge with #953

Issue #847, opened 2016-03-30
Support u alias for url_encode


alethea:
Currently, using the shorthand [`.u` ](http://ruby-doc.org/stdlib-2.2.4/libdoc/erb/rdoc/ERB/Util.html#method-i-u) alias for [`ERB::Util.url_encode`](http://ruby-doc.org/stdlib-2.2.4/libdoc/erb/rdoc/ERB/Util.html#method-i-url_encode) in views marks escaped model attributes as unsafe XSS, event though `.url_encode` is counted as safe. Would it be possible to include `.u` in the list of ignored XSS methods?

presidentbeef:
Hi Alethea,

Of course, no problem. Thank you for pointing it out.

presidentbeef:
I'm going to see if anyone wants to take this issue on as an easy contribution.

This just involves adding `:u` where `:url_encode` is listed as an ignored method and then adding a couple tests as [documented here](https://github.com/presidentbeef/brakeman/wiki/How-to-Report-a-Brakeman-Issue#submitting-a-pull-request).

PR #863, opened 2016-04-19
Added Support u alias for url_encode

cdollins:
XSS will no longer be warned for `:u` alias of `:url_encode` as per request #847


test/apps/rails2/app/views/home/test_model.html.erb
```diff
@@ -11,3 +11,5 @@ It's just a model <%= link_to "Hipster ipsum", User.first %>
 It's just a couple of models <%= link_to "Hipster ipsum", [Account.first, User.last] %>
 
 <%= render :partial => 'models', :collection => all_the_things, :as => :model %>
+
+Bad for 2.x: <%= link_to "test", params[:user_id].u %>
```
presidentbeef: While this happens to work... `u` would be actually called `u(params[:user_id])` :hamster:

cdollins: I just figured this out. 🐼

presidentbeef:
Thanks! 🚀

PR #864, opened 2016-04-19
Report end lines in Code Climate issues

GordonDiggs:
Our spec requires both beginning and ending lines, so this should
include both, even if they are the same.

Spec reference:
https://github.com/codeclimate/spec/blob/master/SPEC.md#locations


presidentbeef:
🆗

GordonDiggs:
Thanks @presidentbeef!

PR #868, opened 2016-04-30
Skip branches with `Rails.env.test?'

### LINKED ISSUE #862: ignore code inside of 'if Rails.env.test?'
we have a few hacks here and there that bypass authentication / add catch-all routes to make tests easier to write ... would be nice if those would not get caught as errors

is that possible / would that make sense ?

@presidentbeef

presidentbeef:
closes #862 

@grosser can you test and see if this does what you wanted?

### DIFF
--- a/lib/brakeman/processors/alias_processor.rb
+++ b/lib/brakeman/processors/alias_processor.rb
@@ -87,6 +87,7 @@ class Brakeman::AliasProcessor < Brakeman::SexpProcessor
 
   ARRAY_CONST = s(:const, :Array)
   HASH_CONST = s(:const, :Hash)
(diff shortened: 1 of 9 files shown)


grosser:
tried adding a match all route as a test and it was a warning:

```
# config/routes.rb
  if Rails.env.test?
    match '/:controller/:action'
  end
```

```
High ... routes.rb near line 124 |
```

presidentbeef:
Okay, thanks. The specific test case is helpful.

presidentbeef:
Should work now.

grosser:
confirmed working! :D

presidentbeef:
Great, thank you for testing!

PR #872, opened 2016-05-06
Support multiple newlines between ruby expressions in ERB files

ptoomey3:
We ran into a recurring issue where line numbers within ERB files were incorrect. They were usually one or two lines off, but it wasn't super clear why. Upon looking at the issue it seems that Brakeman's overrides in erubis for handling newlines is flawed. The conditional has no branch that handles more than a single newline correctly. If you have a run of several newlines such as `\n\n\n\n` the old logic would process it like this:

* does `text == "\n"`? - Nope
* elsif `text.include? "\n"`? - Yup

But, the issue is that when that branch is taken, we end up with this:

``` ruby
text = "\n\n\n\n"
lines = text.split("\n")
=> []
```

As a result, there is nothing to iterate over inside that branch. This is semi-unexpected (at least by me) since adding any non-newline character completely changes that split.

``` ruby
text = "\n\n\n\n "
lines = text.split("\n")
=> ["", "", "", "", " "]
```

So, we need to handle runs of newlines explicitly. Luckily we can just change the original `text == "\n"` to a regex for one or more newlines. 

/cc @gregose

### DIFF
--- a/lib/brakeman/parsers/rails3_erubis.rb
+++ b/lib/brakeman/parsers/rails3_erubis.rb
@@ -9,8 +9,8 @@ class Brakeman::Rails3Erubis < ::Erubis::Eruby
 
   #This is different from Rails 3 - fixes some line number issues
   def add_text(src, text)
-    if text == "\n"
-      src << "\n"
+    if text =~ /\A\n+\z/
+      src << text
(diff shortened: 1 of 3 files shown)


presidentbeef:
Great, thank you!

PR #873, opened 2016-05-09
More ERB newline fixes

ptoomey3:
In https://github.com/presidentbeef/brakeman/pull/872 I fixed an issue where there were multiple newlines in between ERB expressions. But, after merging that, I noticed that the existing code also failed to properly handle multiple newlines in between HTML markup and ruby code. This seems to have been a known issue, as a number of unit tests were annotated with something like "Should be line 4 instead of 3". The existing code failed to handle anything like this:

```
<h1>foo</h1>

<% if @bar%>
```

The existing regex parsing logic would result in a scan match with a `text` value of `<h1>foo</h1>\n\n`. So, we have a string that has two trailing newlines. But, when you do `lines.split("\n")` you end up with `["<h1>foo</h1>"]`. There was some exiting logic to handle trailing newlines:

``` ruby
if text.match(/\n\z/)
  ...
```

But, the above code only handles the scenario where there is a single trailing newline. So, rather than make another similar patch like I did prior, I've refactored this bit of logic to be a bit more general purpose with regard to newlines. 

/cc @gregose


lib/brakeman/parsers/rails3_erubis.rb
```diff
@@ -9,23 +9,14 @@ def add_preamble(src)
...
   def add_text(src, text)
-    if text =~ /\A\n+\z/
-      src << text
-    elsif text.include? "\n"
-      lines = text.split("\n")
-      if text.match(/\n\z/)
-        lines.each do |line|
-          src << "@output_buffer << ('" << escape_text(line) << "'.html_safe!);\n"
-        end
+    lines = text.lines
```
ptoomey3: Unlike "text.split("\n")`, the `String#lines` method will split things in a predictable way across newlines.

test/tests/rails3.rb
```diff
@@ -653,7 +663,7 @@ def test_xss_iteration2
   def test_unescaped_model
     assert_warning :type => :template,
       :warning_type => "Cross Site Scripting",
-      :line => 3, #This should be line 4 :(
```
ptoomey3: Yay! A bunch of these were able to be fixed up.

test/tests/rails3.rb
```diff
@@ -694,7 +704,7 @@ def test_sql_injection_in_template
     #SQL injection in controllers should not warn again in views
     assert_no_warning :type => :template,
       :warning_type => "SQL Injection",
-      :line => 3, #This should be line 4 :(
```
ptoomey3: I changed this one, since it looked like it would have been caused by the same issue. But, I wasn't 100% sure, since this test passed either way, given it is a `assert_no_warning`.

presidentbeef:
At some point perhaps this should just be re-synced with Rails' implementation. Not sure how much that will break though.

presidentbeef:
Thank you again :)

PR #878, opened 2016-05-11
Sync ERB handling up with rails/rails from Rails 5

ptoomey3:
Per your comment in https://github.com/presidentbeef/brakeman/pull/873#issuecomment-217929880, this PR syncs up ERB handling with Rails 5 from upstream. The one caveat here is that upstream uses `@output_buffer.safe_expr_append` during code generation. Given this is all being analyzed statically, I'm not sure that matters or not if we are analyzing a rails 3 app. Since it is all just an intermediate format used for analysis anyway, I tend to think it doesn't matter. But, I'll leave it up to you if you would like to override that change from upstream or not.

### DIFF
--- a/lib/brakeman/parsers/rails3_erubis.rb
+++ b/lib/brakeman/parsers/rails3_erubis.rb
@@ -1,45 +1,74 @@
 Brakeman.load_brakeman_dependency 'erubis'
 
-#This is from Rails 3 version of the Erubis handler
+# This is from Rails 5 version of the Erubis handler
+# https://github.com/rails/rails/blob/ec608107801b1e505db03ba76bae4a326a5804ca/actionview/lib/action_view/template/handlers/erb.rb#L7-L73
 class Brakeman::Rails3Erubis < ::Erubis::Eruby
 
   def add_preamble(src)
-    # src << "_buf = ActionView::SafeBuffer.new;\n"
+    @newline_pending = 0
+    src << "@output_buffer = output_buffer || ActionView::OutputBuffer.new;"
   end
 
-  #This is different from Rails 3 - fixes some line number issues
   def add_text(src, text)
-    lines = text.lines
-    lines.each do |line|
-      if line == "\n"
-        src << line
-      else
-        newline = line.end_with?("\n") ? "\n" : nil
-        src << "@output_buffer << ('" << escape_text(line.chomp) << "'.html_safe!);#{newline}"
-      end
+    return if text.empty?
+
+    if text == "\n"
+      @newline_pending += 1
+    else
+      src << "@output_buffer.safe_append='"
+      src << "\n" * @newline_pending if @newline_pending > 0
+      src << escape_text(text)
+      src << "'.freeze;"
+
+      @newline_pending = 0
+    end
+  end
+
+  # Erubis toggles <%= and <%== behavior when escaping is enabled.
+  # We override to always treat <%== as escaped.
+  def add_expr(src, code, indicator)
+    case indicator
+    when '=='
+      add_expr_escaped(src, code)
+    else
+      super
     end
   end
 
   BLOCK_EXPR = /\s*((\s+|\))do|\{)(\s*\|[^|]*\|)?\s*\Z/
 
   def add_expr_literal(src, code)
+    flush_newline_if_pending(src)
     if code =~ BLOCK_EXPR
       src << '@output_buffer.append= ' << code
     else
-      src << '@output_buffer.append= (' << code << ');'
+      src << '@output_buffer.append=(' << code << ');'
     end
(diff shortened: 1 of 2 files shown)


lib/brakeman/parsers/rails3_erubis.rb
```diff
@@ -1,45 +1,74 @@
...
-      src << '@output_buffer.append= (' << code << ');'
+      src << '@output_buffer.append=(' << code << ');'
     end
   end
 
   def add_expr_escaped(src, code)
+    flush_newline_if_pending(src)
     if code =~ BLOCK_EXPR
-      src << "@output_buffer.safe_append= " << code
+      src << "@output_buffer.safe_expr_append= " << code
```
ptoomey3: Here is the change from `safe_append=` to `safe_expr_append=`. I tend to think it is easiest (and safe) to just match upstream.

lib/brakeman/processors/erubis_template_processor.rb
```diff
@@ -2,7 +2,7 @@
 
 #Processes ERB templates using Erubis instead of erb.
 class Brakeman::ErubisTemplateProcessor < Brakeman::TemplateProcessor
-  
+
```
ptoomey3: Unless you are opposed, I'm going to stop undoing the whitespace fixes that happen when I click save in my editor. Maybe over time they will all be fixed 😛. But seriously, if you would prefer me to rip out whitespace changes to preserve the blame log just let me know.

.gitignore
```diff
@@ -3,3 +3,4 @@ coverage/
 test/coverage/
 .bundle
 bundle
+.tags
```
ptoomey3: I kept running into this and finally forgot to remove my tags file before pushing. Let me know if you prefer this in a unique PR.

lib/brakeman/processors/erubis_template_processor.rb
presidentbeef: No worries. I tend to fix them as I go, too.

.gitignore
presidentbeef: This seems pretty specific to your setup...

ptoomey3: I can rip it out...It is generated by whatever some atom plugin is kicking of ctags with. I just added it to my global .gitignore.

presidentbeef: Yes, please.

ptoomey3: Ok...removed and squashed commit history.

lib/brakeman/parsers/rails3_erubis.rb
```diff
@@ -1,45 +1,74 @@
...
 
-#This is from Rails 3 version of the Erubis handler
+# This is from Rails 5 version of the Erubis handler
+# https://github.com/rails/rails/blob/ec608107801b1e505db03ba76bae4a326a5804ca/actionview/lib/action_view/template/handlers/erb.rb#L7-L73
 class Brakeman::Rails3Erubis < ::Erubis::Eruby
 
   def add_preamble(src)
-    # src << "_buf = ActionView::SafeBuffer.new;\n"
+    @newline_pending = 0
+    src << "@output_buffer = output_buffer || ActionView::OutputBuffer.new;"
```
presidentbeef: Just as a note, this line and the postamble will be ignored anyway. Otherwise this causes a lot of overhead processing templates.

ptoomey3: Are you cool with leaving the code so as to match upstream rails?

presidentbeef: Yeah, it's okay. More of a note for the future.

gregose:
FWIW, this also fixes an issue with `Rails3Erubis` parsing of case statements:

Before the fix:

```
=> [{:error=>"/test_app/app/views/log/_action.html.erb:2 :: parse error on value \"@output_buffer\" (tIVAR)",
  :location=>"Could not parse /test_app/app/views/log/_action.html.erb"}]
```

This is due to prepended `@output_buffer` in the generated ruby source:

```ruby
case view.action
@output_buffer << (''.html_safe!); when "foo" ;@output_buffer.append= ( "bar" );
@output_buffer << (''.html_safe!); when "foo2" ;@output_buffer.append= ( "bar2" );
end
@output_buffer << (''.html_safe!);
```

After the fix to move to the upstream ERB parsing:

```ruby
@output_buffer = output_buffer || ActionView::OutputBuffer.new; case view.action
 when "foo" ;@output_buffer.append=( "bar" );@output_buffer.safe_append='
'.freeze; when "foo2" ;@output_buffer.append=( "bar2" );@output_buffer.safe_append='
'.freeze; end
@output_buffer.to_s
```

PR #879, opened 2016-05-11
Add ability to bypass the "is this a rails app" check

oreoshake:
I would not call this "sinatra support", but at least we can get results for junk in lib. This change caught a bug in a non-rails app.

/to @presidentbeef 
/cc @ptoomey3 @gregose

### DIFF
--- a/lib/brakeman/options.rb
+++ b/lib/brakeman/options.rb
@@ -276,6 +276,10 @@ module Brakeman::Options
           options[:show_version] = true
         end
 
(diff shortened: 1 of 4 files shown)


bf4:
What bug?

oreoshake:
By "bug" I mean vulnerability. It would have caught an obvious RCE in a Sinatra app. Sure it had a few false positives, but that's chill.

presidentbeef:
Dangerous :warning:

Issue #880, opened 2016-05-13
A Collection's methods are keyed regardless of level


imtayadeway:
When, say, a model's methods are indexed, instance and class level methods are keyed in the same way. So if an instance and a class level method share the same name, one is lost.

### Example
Consider the following model:
```ruby
class Foo < ActiveRecord::Base
  def self.bar(something)
    eval(something)
  end

  def self.baz(something)
    eval(something)
  end

  def bar(something)
    self.class.bar(something)
  end
end
```

As the model gets processed, its representation in `Brakeman::Model` will receive `#add_method` 3 times, one for each method defined above. Here's what happens to the index on each call:

1.
```ruby
@methods # => {:public=>{:bar=>{:src=>s(:defs, :Foo, :bar, s(:args, :something), s(:call, nil, :eval, s(:lvar, :something))), :file=>"/home/tim/src/dummy/app/models/foo.rb"}}, :private=>{}, :protected=>{}}
```
2.
```ruby
@methods # =>  {:public=>
  {:bar=>{:src=>s(:defs, :Foo, :bar, s(:args, :something), s(:call, nil, :eval, s(:lvar, :something))), :file=>"/home/tim/src/dummy/app/models/foo.rb"},
   :baz=>{:src=>s(:defs, :Foo, :baz, s(:args, :something), s(:call, nil, :eval, s(:lvar, :something))), :file=>"/home/tim/src/dummy/app/models/foo.rb"}},
 :private=>{},
 :protected=>{}}
```
3.
```ruby
@methods # => {:public=>
  {:bar=>{:src=>s(:defn, :bar, s(:args, :something), s(:call, s(:call, s(:self), :class), :bar, s(:lvar, :something))), :file=>"/home/tim/src/dummy/app/models/foo.rb"},
   :baz=>{:src=>s(:defs, :Foo, :baz, s(:args, :something), s(:call, nil, :eval, s(:lvar, :something))), :file=>"/home/tim/src/dummy/app/models/foo.rb"}},
 :private=>{},
 :protected=>{}}
```

As you can see, on the third iteration the previously defined `:bar` (class method) has been replaced by the `:bar` instance method.

### Consequences

This has caused an existing security issue to go unnoticed. It was shown that merely renaming one of the methods, or transposing their order would cause Brakeman to correctly raise a warning.

I'm happy to help tackle this issue if needed. Thanks!

PR #881, opened 2016-05-16
Key instance and class methods differently

imtayadeway:
If a collection had a class method and an instance method that shared
the same name, one would overwrite the other since they are keyed by
name only.

This change keys the class methods prefixed with this class. Since the
tracker does this in a couple of different places anyway, it can now
just rely on the key to return the method name.

Since fingerprints are based partly on this key, a few tests had to be
updated that checked for this.

Fixes https://github.com/presidentbeef/brakeman/issues/880

### DIFF
--- a/test/tests/rails5.rb
+++ b/test/tests/rails5.rb
@@ -13,7 +13,7 @@ class Rails5Tests < Test::Unit::TestCase
       :controller => 0,
       :model => 0,
       :template => 2,
-      :generic => 6
+      :generic => 7
     }
   end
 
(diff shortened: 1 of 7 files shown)


imtayadeway:
I'm a little unfamiliar with the layout of the test suite, so haven't added a regression test for this yet. Should one just go under the Rails 5 app? I will happily do so with a point in the right direction. Thanks!

presidentbeef:
Hi Tim,

Thank you for the pull request.

Yes, please add a test case to the "Rails 5" app. Ideally with tests that include the `:method ` value.

imtayadeway:
@presidentbeef test added.  Here's a screenshot demonstrating that it fails on master:

screenshot from 2016-05-18 11 50 23

test/tests/rails5.rb
```diff
@@ -206,4 +206,16 @@ def test_cross_site_scripting_strip_tags_cve
...
+  def test_dangerous_eval_in_prior_class_method_with_same_name
+    assert_warning :type => :warning,
+      :warning_code => 13,
+      :fingerprint => "7fe3142d1d11b7118463e45a82b4b7a2b5b5bac95cf8904050c101fae16b8168",
+      :warning_type => "Dangerous Eval",
+      :line => 3,
+      :message => /User input in eval near line 3/,
+      :confidence => 0,
+      :relative_path => "app/models/user.rb",
+      :user_input => s(:params)
```
presidentbeef: Very minor thing. Can you please add

    :method => :"User.evaluate_user_input"

to be explicit

imtayadeway: Fixed! I had noted that you had asked for that above, but then completely forgot :blush:

presidentbeef:
Thank you!

PR #885, opened 2016-05-24
Allow extended regexp into validate format

c0va23:
Hi!

I just want to apologize for my English. I use Google Translate for this message.

I add support of extended regexes into checking of validate format.

I like to write complex regular expressions in a multiline format, but the brakeman displays false warnings on these regular expressions.

In tests I've added an example of such a regular expression.

### DIFF
--- a/lib/brakeman/checks/check_validation_regex.rb
+++ b/lib/brakeman/checks/check_validation_regex.rb
@@ -59,17 +59,37 @@ class Brakeman::CheckValidationRegex < Brakeman::BaseCheck
     end
   end
 
+  # Match secure regexp without extended option
+  SECURE_REGEXP_PATTERN = %r{
+    \A
+    \\A
+    .*
+    \\[zZ]
+    \z
+  }x
+
+  # Match secure of regexp with extended option
+  EXTENDED_SECURE_REGEXP_PATTERN = %r{
+    \A
+    \s*
+    \\A
+    .*
+    \\[zZ]
+    \s*
+    \z
+  }mx
+
   #Issue warning if the regular expression does not use
   #+\A+ and +\z+
   def check_regex value, validator
     return unless regexp? value
 
-    regex = value.value.inspect
-    unless regex =~ /\A\/\\A.*\\(z|Z)\/(m|i|x|n|e|u|s|o)*\z/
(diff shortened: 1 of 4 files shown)


lib/brakeman/checks/check_validation_regex.rb
```diff
@@ -59,17 +59,37 @@ def process_validates validator
     end
   end
 
+  # Match secure regexp without extended option
+  SECURE_REGEXP_PATTER = %r{
+    \A
+    \\A
+    .*
+    \\[z|Z]
```
presidentbeef: This is not the same as `(z|Z)`. User either `(z|Z)` or `[zZ]`:

```
2.3.0 :001 > "|".match /[z|Z]/
 => #<MatchData "|">
2.3.0 :002 > "|".match(/(z|Z)/)
 => nil
```

lib/brakeman/checks/check_validation_regex.rb
```diff
@@ -59,17 +59,37 @@ def process_validates validator
     end
   end
 
+  # Match secure regexp without extended option
+  SECURE_REGEXP_PATTER = %r{
```
presidentbeef: Please add the `N` on the end of this constant and the other one.

lib/brakeman/checks/check_validation_regex.rb
```diff
@@ -59,17 +59,37 @@ def process_validates validator
...
+    \A
+    \\A
+    .*
+    \\[z|Z]
+    \z
+  }x
+j
```
presidentbeef: What is this `j` doing here and how are tests passing with it...

test/tests/rails_with_xss_plugin.rb
```diff
@@ -299,6 +299,15 @@ def test_format_validation_21
       :file => /user\.rb/
   end
 
+  def test_format_validation_with_multiline
```
presidentbeef: This should be in `test/tests/rails4.rb` not here

c0va23:
@presidentbeef, i fix all tupos and notes

presidentbeef:
Hi Fedorenko,

Thank you for your contribution and making those changes!

PR #909, opened 2016-07-15
Provide the location brakeman attempted to scan when no app is found.

oreoshake:
This is less useful for CLI users, but when integrating brakeman into things it would be useful to know where brakeman tried to look. 

/cc @presidentbeef


presidentbeef:
:boat:

PR #913, opened 2016-07-26
Add support for in-file-comment whitelisting.

zheei:
The downside of the -I option for interactive warning white-listing is that the ignore file can get big and out of date, also the reason for whitelisting is not easy to parse. This pull request addresses these issues, but of course add some other downsides.

Ideal use case:
You get a brakeman report with some warnings you'd like to fix, you fix them then add "Brakeman safe: #{fingerprint}: #{note}" within 5 lines of the warning (top and bottom). Run the brakeman like:

brakeman -o report.html -i config/brakeman.ignore --ignore-file-only
// 1. has to use .html report format to trigger the scanning for the flag (the commenting line)
// 2. --ignore-file-only: to speed things up, optional.

The .ignore file will be updated, including the note.

You push the code change and/or the ignore file. You or someone else pull the change, and sees the comments and realizes that warning line has been whitelisted.

Downside:
Could not find a easy way to tease out the context scanning functionality of the html report, so basically the brakeman has to run twice: once to fill the .ignore file, another time to take in the ignore file.

Tests:
Minimal.

Documentation:
Minimal.

Please advise.

### DIFF
--- a/lib/brakeman/options.rb
+++ b/lib/brakeman/options.rb
@@ -277,6 +277,10 @@ module Brakeman::Options
           options[:show_version] = true
         end
 
+        opts.on "--ignore-file-only", "Do not generate report, only update ignore file." do
+          options[:ignore_file_only] = true
+        end
+
         opts.on "--force-scan", "Scan application even if rails is not detected" do
           options[:force_scan] = true
         end
--- a/lib/brakeman/report/ignore/config.rb
+++ b/lib/brakeman/report/ignore/config.rb
@@ -103,8 +103,13 @@ module Brakeman
         w
       end.sort_by { |w| w[:fingerprint] }
 
+      # For each white listed entry, we check if it's already in the ignored list, if not,
+      # we write them out.
+      @total_ignored ||= @already_ignored
+      warnings.reject! { |r| @ignored_fingerprints.include? r[:fingerprint] }
+      @total_ignored += warnings
       output = {
-        :ignored_warnings => warnings,
+        :ignored_warnings => @total_ignored,
         :updated => Time.now.to_s,
         :brakeman_version => Brakeman::Version
       }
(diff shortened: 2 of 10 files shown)


presidentbeef:
Hi Zheei,

Thank you for this contribution. However, I don't think this is a useful feature for many users.

My understanding is that you want, in your pull requests, to indicate the code in which a warning is being ignored. Perhaps this would be better in a post-Brakeman tool that pulls out that information?

zheei:
Hi presidentbeef:

Thank you for checking this out. I'm working with a fairly sized Rails code base with lots moving parts, and we'd like to keep things as tight as possible. The Brakeman is great for scanning and failing build for our build process, but due to the size and fragmentation of our code base we'd like the ignored warnings to be more visible in the code base itself and to not having the developers worrying about maintaining the ignore file. Our ignore file size prior is 200+MB and rising, and now it's an empty template. For the visibility issue, we thought about a post-Brakeman tool, but it still does not beat having a note right near where the warning is in the code, and it would be quite some work to edit the files to add the notes in bulk with another tool.

presidentbeef:
I understand your pain. However, I would really prefer to only support one method for ignoring warnings and the central ignore file was the decision.

There are other options for reducing the number of warnings if you have lots of false positives of the same kind: http://brakemanscanner.org/docs/reducing_false_positives/

Would it also be helpful to prune the ignore file? That is a planned feature.

zheei:
Thanks for the pointers, we did check out the options for limiting false-positive, though it turns out to be very effective for us, and we use -w3 when we fail builds.

I do think pruning the ignore file will be helpful, if by pruning you mean delete the ones that's no longer pointing to real warnings. I'm glad that's a planned feature.

I was not very opportunistic on the pull request getting accepted because it does smell like a hack. But it worked for us so far, so figured might as well put it out there. Thanks again.

PR #915, opened 2016-07-27
Fix gem building for Code Climate engine

presidentbeef:
The gemspec was refactored a bit with the `3.3.1` release and that broke the Docker configuration.

I think this fixes it, although Docker is broken on my machine so at the moment I can only confirm the engine builds successfully.

/cc @GordonDiggs


GordonDiggs:
@presidentbeef I just tested this locally and it seems to be working! Thanks for pinging me.

presidentbeef:
Great, thank you for confirming.

PR #935, opened 2016-09-06
Add new plain text output format

presidentbeef:
This will eventually be the default output format. Suggestions welcome.

image


grosser:
looks good!

dgsprb:
Nice!

costajohnt:
was more readable.  thanks for this.

PR #936, opened 2016-09-07
Use Correct Term "Cross-Site"

paultetreau:
Replaced all instances of "cross site" with the properly hyphenated "cross-site", preserving case.


b264:
:shipit:

presidentbeef:
Hi Paul,

Thank you for the suggestion. However, it is not possible to change all instances of the term without essentially breaking out external API. Perhaps this could be a part of Brakeman 4.0, but the change cannot be made before then.

PR #937, opened 2016-09-08
Add support for Code Climate debug option

ABaldwinHunter:
cc @mrb

### DIFF
--- a/bin/codeclimate-brakeman
+++ b/bin/codeclimate-brakeman
@@ -12,6 +12,13 @@ end
 
 command = "/usr/src/app/bin/brakeman --quiet --format codeclimate"
 
+debug = engine_config["config"] && engine_config["config"]["debug"]
+
+# ATM this gets parsed as a string instead of bool: "config":{ "debug":"true" }
+if debug && debug.to_s.downcase == "true"
+  command << " --debug"
+end
+
 if engine_config["include_paths"]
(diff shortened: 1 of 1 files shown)


presidentbeef:
This won't affect the Code Climate report. Will you be capturing STDERR?

presidentbeef:
@ABaldwinHunter any comment? Just trying to understand the use case.

ABaldwinHunter:
Hey @presidentbeef! apologies missed your earlier comment. Thanks for following up!

We capture and output stderr in two cases:

- [in the CLI](https://github.com/codeclimate/codeclimate#environment-variables) when `CODECLIMATE_DEBUG=1` is passed
- on codeclimate.com the stderr gets output occasionally in engine build step, like the duplication engine [here](https://codeclimate.com/github/presidentbeef/brakeman/builds/71)

Typically we wouldn't run engines in that mode, but occasionally the introspection can be useful when an analysis errors.

Let me know what you think :)

presidentbeef:
Makes sense, just wanted to be sure.

I would recommend adding `--no-progress` to avoid control codes in the output. It's a little easier to read that way.

presidentbeef:
Looks like the build failed because `terminal-table` 1.7.2 accidentally broke Ruby 1.9.3 compatibility (https://github.com/tj/terminal-table/issues/79), so no worries.

ABaldwinHunter:
Cool, thanks!

PR #951, opened 2016-10-17
Update ruby_parser gem

soutaro:
Updating to 3.8.3 allows brakeman to parse and analyze Rails apps with imaginary and rational literals. (I have one Rails app with rational literal!)

* Announcement: http://www.zenspider.com/releases/2016/10/ruby_parser-version-3-8-3-has-been-released.html
* Changes from 3.8.1: https://github.com/seattlerb/ruby_parser/blob/master/History.txt#L1-L19

### DIFF
--- a/gem_common.rb
+++ b/gem_common.rb
@@ -5,7 +5,7 @@ module Brakeman
     end
 
     def self.base_dependencies spec
-      spec.add_dependency "ruby_parser", "~>3.8.1"
+      spec.add_dependency "ruby_parser", "~>3.8.3"
       spec.add_dependency "ruby2ruby", "~>2.3.0"
       spec.add_dependency "safe_yaml", ">= 1.0"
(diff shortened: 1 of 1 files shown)


presidentbeef:
Hi Soutaro,

Note that Brakeman ships with whatever version of RubyParser was most recent at the time. So Brakeman 3.4.0 uses RubyParser 3.8.2. There's no real reason to force 3.8.3, although I don't mind merging it. If you install `brakeman-lib` (which does not bundle dependencies) your app should pick up RubyParser 3.8.3 since it still meets the `~>3.8.1` specification (in case you don't want to wait for Brakeman 3.4.1).

soutaro:
Oops... you are right. I'm sorry for unnecessary pull-request. (I was confusing on the `~>` syntax.)

Thanks!

PR #953, opened 2016-10-27
Check forgery setting in all direct subclasses of ActionController::Base (updated)

presidentbeef:
Fixed up version of #858 from @jsyeo


codelion:
Great work @jsyeo!

PR #957, opened 2016-11-07
Raise errors for files with commas in their name

maxjacobson:
We received a report that analysis was failing for a repo and tracked it
down to the fact that there was a file in the repo whose filename ended
in a comma. This leads to a fairly confusing error.

This change surfaces an error which explains to the user what is going
on, and what they can do about it.

If there is a comma in the middle of the filename, that doesn't lead to
a runtime error currently, but it does lead to misbehavior, so I think
an error is also appropriate for those files.

### DIFF
--- a/bin/codeclimate-brakeman
+++ b/bin/codeclimate-brakeman
@@ -19,8 +19,17 @@ if debug && debug.to_s.downcase == "true"
   command << " --debug --no-progress"
 end
 
+IncompatibleFilenameError = Class.new(StandardError)
+
 if engine_config["include_paths"]
-  command << " --only-files #{engine_config["include_paths"].compact.join(",").shellescape}"
+  paths = engine_config["include_paths"].compact.select do |path|
+    if path.include?(",")
+      raise IncompatibleFilenameError, "Can't analyze #{path.inspect} because it has a comma in its filename. Please remove this file or exclude it in your .codeclimate.yml"
+    else
+      path
+    end
+  end
+  command << " --only-files #{paths.join(",").shellescape}"
 end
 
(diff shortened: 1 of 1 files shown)


presidentbeef:
Hi Max,

Would it make sense to support these (weird) file names?

maxjacobson:
Hey Justin! Hope you're doing well.

In a perfect world, yes, but IMO it's not necessary here in our world 😄.

If you want to support it, the one way I've thought of that would work without breaking anything with regard to backwards compatibility is to introduce a new command line flag that lets you add an individual file rather than a list of files. It could be used in conjunction with the existing one:

```ruby
#!/usr/bin/env ruby

require 'optparse'
require 'set'

options = {}
opts = OptionParser.new

opts.on "--only-files file1,path2,etc", Array, "Process only these..." do |files|
  options[:only_files] ||= Set.new
  options[:only_files].merge files
end

opts.on "--individual-only-files file1", String, "Process only this..." do |file|
  options[:only_files] ||= Set.new
  options[:only_files].merge [file]
end

opts.parse!

p options
p ARGV
```

```
$ ./inputter --only-files foo,bar,baz --only-files quack\,quack --individual-only-files bar,baz
{:only_files=>#<Set: {"foo", "bar", "baz", "quack", "bar,baz"}>}
[]
```

It is possible to make a custom converter to manually parse the input rather than simply doing `.split(",")` which is what the built-in array converter seems to do... but I'm not sure what we would actually do there instead

What do you think?

presidentbeef:
@maxjacobson how about #960 instead?

maxjacobson:
Closing in favor of #960

PR #958, opened 2016-11-07
Warning when `protected_from_forgery` is mixed in.

louim:
Hi! since https://github.com/presidentbeef/brakeman/pull/953 was merged in, we started to see `'protect_from_forgery' should be called in ...` in our controllers that were not inheriting from the `ApplicationController`. 

I had a look, and it seems that it's because we are mixin in the `protect_form_forgery` in those controllers. I added a failing test case that should normally not generate a warning. My knowledge of the Brakeman codebase was not enough for me to create a fix.

Let me know if I can do anything else to help!


presidentbeef:
Hi Louis-Michel,

Thank you for the nice test case! I will see about fixing this.

PR #960, opened 2016-11-10
Update Code Climate to use API instead of CLI

presidentbeef:
Avoid trying to build up a command line invocation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
-script: "bundle exec ruby test/test.rb"
+script:
+  - "bundle exec ruby test/test.rb"
+  - "bundle exec codeclimate-test-reporter"
 
 branches:
   only:
--- a/bin/codeclimate-brakeman
+++ b/bin/codeclimate-brakeman
@@ -1,8 +1,14 @@
 #!/usr/bin/env ruby
 $:.unshift "#{File.expand_path(File.dirname(__FILE__))}/../lib"
 
+require "brakeman"
 require "json"
-require "shellwords"
+
+options = {
+  :output_format => :codeclimate,
+  :quiet => true,
+  :app_path => Dir.pwd
+}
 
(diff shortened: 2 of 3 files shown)


maxjacobson:
Ah, this approach seems really solid! Do you want me to try building and testing the engine on a repo with some strangely named files tomorrow?

presidentbeef:
@maxjacobson Yes, please give it a spin.

maxjacobson:
OK! I can confirm that this does neatly work around the issue in the case where there is a file in the root of the project called, e.g. `foo,`. No more error, and it does include the file.

One other thing I'm observing at this point is that if I rename an existing file, e.g. `app/views/codes/show.html.slim` -> `app/views/codes/show,.html.slim`, then it just silently skips that file without raising an error, even on this branch. I confirmed that it is present in the [`template_paths`](https://github.com/presidentbeef/brakeman/blob/142106b015f414e21d05cb2b1c011ca95a53ab6f/lib/brakeman/app_tree.rb#L103-L105), so I'm not sure where it's being filtered out

presidentbeef:
@maxjacobson does the file name with a comma get skipped even without the `include_paths` option?

maxjacobson:
@presidentbeef Correct. Commented out those lines and the behavior is the same. The paths aren't passed, and the options (when parsed) look like this:

```
{:quiet=>true,:output_format=>:to_codeclimate, :output_files=>["/dev/stdout"]}
```

And it skips the file called `app/views/codes/show,.html.slim`.

This is the same on master as this branch. So, to be clear, this branch looks good to me 👍 but it seems like the AppTree logic might skip some (strangely named) files fwiw

presidentbeef:
Yeah, that's what I suspected. I think I'm okay with that...

maxjacobson:
👍 same

PR #964, opened 2016-11-17
Change warnings link in README

pocke:
`WARNING_TYPES` seems not to be maintained already.
So, I've removed it. And changed a link in README to `doc/warning_types/`


presidentbeef:
Yes, that makes sense. Thanks!

PR #974, opened 2016-12-14
check for the latest version with new option

tamgrosser:
wip @grosser

### DIFF
--- a/bin/brakeman
+++ b/bin/brakeman
@@ -57,6 +57,13 @@ if options[:quiet].nil?
 end
 
 begin
+  if options[:ensure_latest]
+    if error = Brakeman.ensure_latest
+      warn error
+      exit Brakeman::Not_Latest_Version_Exit_Code
+    end
+  end
+
   if options[:previous_results_json]
     require 'json'
     vulns = Brakeman.compare options.merge(:quiet => options[:quiet])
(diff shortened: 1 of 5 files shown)


presidentbeef:
Hi,

Thank you for putting this together. I think it would be better to use the `Gem` interface though, in case of API changes: http://www.rubydoc.info/stdlib/rubygems/Gem.latest_version_for

grosser:
update!
a few observations:
 - tests are in a weird folder and do not end in _test.rb ... makes finding/working with them hard
 - tests cannot run on their own ... makes testing in isolation hard
 - there are no tests for commandline option handling in bin/brakeman

presidentbeef:
Thanks, this looks good.

* The tests are in `test/tests/` and used to also have `test_...` in the name and I couldn't take the triple redundancy.
* You can run individual tests with `ruby test/test.rb --name ...`
* Yes, this is true and has come up before and someone said they'd contribute testing for it...

grosser:
Anything needed for merge ?

presidentbeef:
Seems to work fine.

A couple minor things:

* This should have a specific exit code like the `-z`
* The option should be up around `-z` not in the scanning options section

grosser:
updated, looks better ?

presidentbeef:
Thanks!

PR #975, opened 2016-12-16
normalize test suite / make each test runnable / always use bundler

grosser:
@presidentbeef 

this makes this project not have a wtf test setup :D 
... more things to do, but the first rough cut ...
also using minitest-rg to get clear / non-rainbow test output


grosser:
all green!

grosser:
@presidentbeef merge ?

presidentbeef:
Not a fan of these changes.

I don't want to be forced to use Bundler or Rake or Simplecov. The current Rake setup should be fine.

Moving the tests to `test/unit/` doesn't make sense because most of the tests aren't unit tests.

This adds a dependency to have red/green output instead of rainbow output. Why do you hate rainbows? :rainbow:

If you want to run individual tests, you can use Minitest's `--name` option. E.g., ` ruby test/test.rb --name /Rails4Tests/` or `ruby test/test.rb --name test_relative_path_in_warnings`

grosser:
please reopen.
 - rainbow is distracting and gives no clear indication if the test failed
or succeeded
 - not forcing bundler means funky/random errors because we pick random
dependencies ... can add a "NO_BUNDLER" option if you want to use it
without ...
 - ruby test/test.rb does not work, that's what I meant with individual
tests

grosser:
pick the one you find more readable:

screen shot 2017-04-21 at 8 51 08 am

mltsy:
Awwww, not the rainbows!  Hey, Github is a free country, you can always make your own non-rainbow dev fork ;)  (Bundler is a good idea though, and I kinda like adding simplecov too, you should try it! Or do you not like it for some reason?)

grosser:
prefer https://github.com/grosser/single_cov ... forces you to write good unit tests and give you actionable/fast feedback on coverage

mltsy:
Oh, I was asking @presidentbeef about why he didn't like simplecov actually, but it appears neither of you like it, haha!  I haven't tried single_cov, but having some kind of coverage report is a nice automatic improvement either way I think.  (Just my opinion though - I'm not the one who maintains the gem :wink:)

grosser:
yeah, any kind of coverage alert would be nice ... I doubt all PRs add
coverage for all the code they add

presidentbeef:
I like simplecov, it's run for Brakeman if it's installed:https://github.com/presidentbeef/brakeman/blob/master/test/test.rb#L5-L11

I don't like it to run all the time though because it generates a bunch of files in CWD that I have to clean up.

mltsy:
Ooh, so you just let CI run it?  Smart 'nuff! :smile:  Is there a way to get a warning on PRs where coverage goes down (to address @grosser's point)?

grosser:
single_cov can report coverage percent and fail when it goes below X ...
just set X to the current percentage ...

presidentbeef:
@mltsy it's not running in CI at the moment, actually. I generally look at CodeClimate (https://codeclimate.com/github/presidentbeef/brakeman/coverage) for test coverage.

If someone wants to set it up code coverage as part of the tests (such that it doesn't generate a bunch of report files) I'll accept the PR.

mltsy:
Right, that's where I saw it as well.  I guess I was including CodeClimate in "CI" (I don't know if that's technically correct or not?).  I assumed that was generated from simplecov - maybe not?  Either way I thought perhaps there was a way to use the CodeClimate/Github integration to alert of code coverage drops, but I don't see a way in their docs I guess so maybe not!

Issue #910, opened 2016-07-20
Fail on "parse error" ?


grosser:
Just saw that brakeman shows parse errors, but does not fail ... which I would have expected ...

```
  has_many :attachments,
    as: :source,
    dependent: :destroy
```

```
| comment.rb:21 :: parse error on value ":" (tCOLON)     | Could not parse /Users/lliu/code/zendesk/zendesk/app/models/events/comment.rb     |
```

and

```
          config_path, reset, project = opts.reverse_merge!(
            config_path: config_file_path,
            reset: false,
            project: 'default'
          ).values_at(:config_path, :reset, :project)

```

with brakeman (3.3.2)

... they are listed as errors ... and -z says exit on warn ... but it should fail on error ... right ?

@presidentbeef

presidentbeef:
Hi Michael,

I'm not sure it should. If Brakeman has a bug, should that fail your build? In general Brakeman does its best to produce a report no matter what.

Also both of those code snippets parse fine for me, so it's probably something else in the file(s).

jcheatham:
@grosser - we've changed the syntax back to hashrocket style and that just pushes the error to later in the file.. maybe some earlier snippet is changing/breaking brakeman's parse state.

grosser:
so I guess add a `--exit-on-error` flag ?

presidentbeef:
I'm not keen on adding a proliferation of exit options.

grosser:
maybe exit-on-warn could fail when there is an error ?

grosser:
this seems to do the trick:

```
class T
public_class_method def self.blah
#[1, 2].each {|i| puts i}
[1, 2].each do |i|
puts i
end
end
end
T.blah
```

presidentbeef:
Closing this issue, as it is already covered here: https://github.com/seattlerb/ruby_parser/issues/213

I do not plan on making any changes to what causes `-z` to return failure or adding new failure options at this time. If you want this kind of behavior (`--exit-on-errors`, `--exit-on-unused-ignore`, etc.) it is pretty easy to write wrapper scripts to do exactly what you want (e.g. [this rake task](https://github.com/presidentbeef/brakeman/blob/dfe03e54fcf170c6a27877f2ecd24002db3f2d54/lib/brakeman/brakeman.rake#L13-L15) although I don't recommend using Rake itself) once #894 is closed out.

grosser:
just ran into this again ... would you accept a PR to add `--exit-on-error` ?

[...]

PR #991, opened 2017-02-04
allow failing on error with --exit-on-error

grosser:
@presidentbeef 

solves https://github.com/presidentbeef/brakeman/issues/910

tested it with samson and works as expected :)
 - does not fail without flag
 - fails with flag
 - does not fail with flag and no errors

added --no-exit-on-error in case failing on error gets to be the default on day ... not sure if 4.0 would be a good milestone for that ...

### DIFF
--- a/bin/brakeman
+++ b/bin/brakeman
@@ -90,6 +90,11 @@ begin
       exit Brakeman::Warnings_Found_Exit_Code
     end
   end
+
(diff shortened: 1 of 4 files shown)


presidentbeef:
Thanks!

grosser:
how about that other PR ;)
https://github.com/presidentbeef/brakeman/pull/975

PR #1024, opened 2017-03-23
fix NoMethodError exception using --compare flag

### LINKED ISSUE #1023: NoMethodError exception using --compare flag
version = 3.6.0

When using the `--compare FILE` option, the following exception occurs (included some of the debug output that preceded the exception for context)

```
Checking users/setup for XSS
Checks finished, collecting results...
Filtering warnings...
[Notice] Using '/Users/anthlam/Projects/TTM/apangea/config/brakeman.ignore' to filter warnings
Generating report...
Comparison saved in 'brakeman-diff-output.json'
/Users/anthlam/.rvm/gems/ruby-2.2.4@apangea/gems/brakeman-3.6.0/bin/brakeman:95:in `<top (required)>': undefined method `options' for nil:NilClass (NoMethodError)
        from /Users/anthlam/.rvm/gems/ruby-2.2.4@apangea/bin/brakeman:22:in `load'
        from /Users/anthlam/.rvm/gems/ruby-2.2.4@apangea/bin/brakeman:22:in `<main>'
        from /Users/anthlam/.rvm/gems/ruby-2.2.4@apangea/bin/ruby_executable_hooks:15:in `eval'
        from /Users/anthlam/.rvm/gems/ruby-2.2.4@apangea/bin/ruby_executable_hooks:15:in `<main>'

```

steps taken:
1. run `brakeman -o brakeman-output.json`
2. run `brakeman --compare brakeman-output.json -o brakeman-diff-output.json`

The compare scan still completes, and the compare output file is successfully generated.

I think this might be caused by the additions for `--exit-on-error`.  The `--compare` flag causes brakeman to enter the `if` block of [this conditional](https://github.com/presidentbeef/brakeman/blob/master/bin/brakeman#L67). `tracker`, which is being used [here](https://github.com/presidentbeef/brakeman/blob/master/bin/brakeman#L95), outside of that conditional, is only set on the `else` block.

seangransee:
Fixes #1023 

In [brakeman#L95](https://github.com/presidentbeef/brakeman/blob/5788041f662b1e4a3d3bc0fecdb0c480cc06afc2/bin/brakeman#L95), `tracker` is used twice outside the block in which it's defined. I think you intended to have this conditional block inside the `else` block immediately above it.

### DIFF
--- a/bin/brakeman
+++ b/bin/brakeman
@@ -89,12 +89,13 @@ begin
     if tracker.options[:exit_on_warn] and not tracker.filtered_warnings.empty?
       exit Brakeman::Warnings_Found_Exit_Code
     end
+
+    #Return error code if --exit-on-error is used and errors were found
+    if tracker.options[:exit_on_error] and tracker.errors.any?
+      exit Brakeman::Errors_Found_Exit_Code
+    end
   end
 
-  #Return error code if --exit-on-error is used and errors were found
-  if tracker.options[:exit_on_error] and tracker.errors.any?
-    exit Brakeman::Errors_Found_Exit_Code
-  end
 rescue Brakeman::NoApplication => e
   warn e.message
   exit Brakeman::No_App_Found_Exit_Code


presidentbeef:
Thanks!

I think this needs to be in both branches... @grosser ?

grosser:
using compare and exit-on-error does not work together ... how about changing `tracker.options[:exit_on_error]` -> `options[:exit_on_error]` ... so it still blows up when used together (instead of silently ignoring it) but does not blow up when just using compare ?

presidentbeef:
Why can't they work together? `--compare` and `--exit-on-warn` do.

grosser:
they can, just needs more work .. is not a trivial bugfix

presidentbeef:
:thinking: copying the condition to both branches doesn't work? Not sure why that would be complicated.

grosser:
it uses the tracker which the other branch does not have

presidentbeef:
AH okay. Then I'm fine with this.

PR #1029, opened 2017-04-13
Removed false positive for redirect_to in Rails 4

### LINKED ISSUE #1028: False positive - Redirect using to_unsafe_hash method
`redirect_to(params.to_unsafe_hash.merge(:only_path => true, :display => nil))`

Brakeman is flagging this as a vulnerability: `Possible unprotected redirect`.
 
I believe this a false positive though, due to the `:only_path` parameter. I have a PR with a possible fix for that. Feel free to close both issue and PR if they do not apply (I am not very proficient with Ruby).

mario-areias:
Fixes #1028 

Not sure if this PR is correct, this is my first contribution to `Breakeman`. Hopefully is good enough to start a discussion and then move it to the right direction.

Rails 4 has changed how to get the hash of an ActionController::Parameter.
If the user needs the all hash parameters it needs to use the method to_unsafe_hash or to_unsafe_h.
Brakeman now recognizes those method as safe as long as they are followed by :only_path

More details: https://github.com/rails/rails/blob/4-2-stable/actionpack/CHANGELOG.md#rails-420-december-20-2014

### DIFF
--- a/lib/brakeman/checks/check_redirect.rb
+++ b/lib/brakeman/checks/check_redirect.rb
@@ -119,6 +119,17 @@ class Brakeman::CheckRedirect < Brakeman::BaseCheck
     return call_has_param(arg, :to_unsafe_hash) || call_has_param(arg, :to_unsafe_h)
   end
 
+  def call_has_param arg, key
+    if call? arg and call? arg.target
+      target = arg.target
+      method = target.method
+
+      node_type? target.target, :params and method == key
+    else
+      false
+    end
+  end
+
   def has_only_path? arg
     if value = hash_access(arg, :only_path)
       return true if true?(value)
--- a/lib/brakeman/checks/check_redirect.rb
+++ b/lib/brakeman/checks/check_redirect.rb
@@ -105,11 +105,23 @@ class Brakeman::CheckRedirect < Brakeman::BaseCheck
     arg = call.first_arg
 
     if hash? arg
-      if value = hash_access(arg, :only_path)
-        return true if true?(value)
-      end
+      return has_only_path? arg
     elsif call? arg and arg.method == :url_for
       return check_url_for(arg)
+    elsif call? arg and hash? arg.first_arg and use_unsafe_hash_method? arg
+      return has_only_path? arg.first_arg
+    end
+
+    false
+  end
+
+  def use_unsafe_hash_method? arg
+    return call_has_param(arg, :to_unsafe_hash) || call_has_param(arg, :to_unsafe_h)
+  end
+
+  def has_only_path? arg
+    if value = hash_access(arg, :only_path)
+      return true if true?(value)
     end
 
     false
--- a/lib/brakeman/util.rb
+++ b/lib/brakeman/util.rb
@@ -129,9 +129,6 @@ module Brakeman::Util
       key = Sexp.new(:call, s(:params), key)
     end
 
-      # p call
-      # p call.find_index(key)
-
     if index = call.find_index(key)
       return true
     end
--- a/lib/brakeman/util.rb
+++ b/lib/brakeman/util.rb
@@ -124,6 +124,21 @@ module Brakeman::Util
     nil
   end
 
+  def call_has_param call, key
+    if key.is_a? Symbol
+      key = Sexp.new(:call, s(:params), key)
+    end
+
+      # p call
+      # p call.find_index(key)
+
+    if index = call.find_index(key)
+      return true
+    end
+
+    false
+  end
+
   #These are never modified
   PARAMS_SEXP = Sexp.new(:params)
   SESSION_SEXP = Sexp.new(:session)
--- a/lib/brakeman/util.rb
+++ b/lib/brakeman/util.rb
@@ -124,18 +124,6 @@ module Brakeman::Util
     nil
   end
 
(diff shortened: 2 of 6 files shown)


lib/brakeman/util.rb
```diff
@@ -124,6 +124,18 @@ def hash_access hash, key
     nil
   end
 
+  def call_has_param call, key
```
presidentbeef: I believe this method is equivalent to

```ruby
call? arg and call? arg.target and node_type? arg.target.target, :params
```

I don't think it should be a general purpose method living in `Util`.

lib/brakeman/checks/check_redirect.rb
```diff
@@ -105,11 +105,23 @@ def only_path? call
...
       return check_url_for(arg)
+    elsif call? arg and hash? arg.first_arg and use_unsafe_hash_method? arg
+      return has_only_path? arg.first_arg
+    end
+
+    false
+  end
+
+  def use_unsafe_hash_method? arg
+    return call_has_param(arg, :to_unsafe_hash) || call_has_param(arg, :to_unsafe_h)
```
presidentbeef: I'd pull the `call_has_params` logic (as noted below) into here.

lib/brakeman/util.rb
mario-areias: Hum. Not sure if I quite understood what you meant. Where do we check if it's calling the function `:to_unsafe_hash` in your example? In my example, I am using the `key` parameter to do that, but I am cannot see something similar in yours. Could you please elaborate a bit so I can proper fix it? :smile:

lib/brakeman/checks/check_redirect.rb
mario-areias: Sure, I will do that.

mario-areias:
Thanks for reviewing this! Unfortunately, I didn't quitie understand the change you asked me to do it. If you could elaborate a bit, it would be great so I can fix it properly!

lib/brakeman/util.rb
presidentbeef: Yes, sorry. You can check with `#method`.

Something like:

```ruby
if call? arg and call? arg.target and node_type? arg.target.target, :params
  arg.target.method == :to_unsafe_hash or arg.target.method == :to_unsafe_h
end
```

This is just an example. It would probably be nicer to use some variables in there to clarify what the different pieces are.

mario-areias: Sorry for the late reply. That's what I tried to do:

```ruby
def call_has_param call, key
    target = arg.target
    method = target.method
    
    return call? arg and call? target and node_type? target.target, :params and method == key
  end
```

But when I run the tests I get this error:

```
ect.rb:126: void value expression (SyntaxError)
... call? arg and call? target and node_type? target.target, :p...
...                               ^
rake aborted!
Command failed with status (1): [ruby test/test.rb --pride...]
```
There are two different `node_type` methods declared in brakeman. One in `Util` and one in `Sexp`. The `Util` one returns a type, but the `Sexp` one doesn't. So I tried to enforce to use the `Util` method by doing this: `Util.node_type?` but I still get the same error.

I also moved the method `call_has_param` method to the `check_redirect` file as you suggested. I have no ideas in how to fix that. Do you have any hints about this issue?

presidentbeef: This is not related to the return value of any method. It's a syntax error because `return` has different precedence rules. Just remove `return`.

mario-areias: Damn. Sorry, such a newbie mistake. I am almost there. That's the method and almost all tests are working. Except by this one: `Rails4Tests#test_zero_errors [test/test.rb:89]:`

The stack trace is this: 

```
Unexpected warning found: [{:error=>"Expected call or attrasgn or safe_call or safe_attrasgn or super or zsuper or result but given s(:const, :User)", :backtrace=>["/home/user/brakeman/lib/brakeman/checks/check_redirect.rb:124:in `call_has_param'", "/home/user/brakeman/lib/brakeman/checks/check_redirect.rb:119:in `use_unsafe_hash_method?'", "/home/user/brakeman/lib/brakeman/checks/check_redirect.rb:111:in `only_path?'", "/home/user/brakeman/lib/brakeman/checks/check_redirect.rb:38:in `process_result'", "/home/user/brakeman/lib/brakeman/checks/check_redirect.rb:27:in `block in run_check'", "/home/user/brakeman/lib/brakeman/checks/check_redirect.rb:26:in `each'", "/home/user/brakeman/lib/brakeman/checks/check_redirect.rb:26:in `run_check'", "/home/user/brakeman/lib/brakeman/checks.rb:195:in `run_a_check'", "/home/user/brakeman/lib/brakeman/checks.rb:131:in `block (2 levels) in actually_run_checks'"]}].
Expected: 0
  Actual: 1
```

Not sure what this really means, but the error first line of error is `method = target.method` from the `call_has_params` method. Don't know how to fix that. 

Thanks for your guidance, btw :)

presidentbeef: That means `target` is not a `call` sexp. Check first with `if call? target`, etc.

mario-areias:
@presidentbeef thanks for your patience. I think now it's good to go!

presidentbeef:
Thank you for *your* patience!

PR #1039, opened 2017-05-01
Use `#hash` as a cache key instead of SHA1 to improve performance

pocke:
By this change, `brakeman` command will be faster around 14%.

------

Currently, `template_cache` key is a SHA1 of a string that is a stringfield sexp.
The stringifying process is very slow. It calls `Sexp#inspect`, ant the method calls itself recursively.

I've replaced it with `Sexp#hash`. This method is faster than `#inspect`.

Benchmarking
=============

```ruby
require 'benchmark'

def dev # Patched brakeman is installed as version 3.6.1.dev
  system 'brakeman _3.6.1.dev_ > /dev/null 2>&1'
end

def master # origin/master is installed as version 3.6.1
  system 'brakeman _3.6.1_ > /dev/null 2>&1'
end

n = 5

Benchmark.bm(16) do |x|
  x.report('dev'){   n.times{dev}}
  x.report('master'){n.times{master}}
end
```

I've tested in gitlab repository.
https://github.com/gitlabhq/gitlabhq

```bash
$ cd path/to/gitlab
$ ruby bench.rb
                       user     system      total        real
dev                0.000000   0.000000 144.030000 (143.694241)
master             0.000000   0.000000 168.080000 (167.729224)
```

### DIFF
--- a/lib/brakeman/processors/lib/render_helper.rb
+++ b/lib/brakeman/processors/lib/render_helper.rb
@@ -70,7 +70,7 @@ module Brakeman::RenderHelper
     #Hash the environment and the source of the template to avoid
     #pointlessly processing templates, which can become prohibitively
     #expensive in terms of time and memory.
-    digest = Digest::SHA1.new.update(template_env.instance_variable_get(:@env).to_a.sort.to_s << name).to_s.to_sym
+    digest = template_env.instance_variable_get(:@env).to_a.sort.push(name).hash
 
     if @tracker.template_cache.include? digest
       #Already processed this template with identical environment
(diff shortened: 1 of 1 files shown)


presidentbeef:
Hi,

I know this is a performance bottleneck, so I would be happy to see improvements here.

However, there seems to be an issue with using `hash`. I'm guessing it's because `Sexp#hash` [caches the hash value](https://github.com/presidentbeef/brakeman/blob/master/lib/ruby_parser/bm_sexp.rb#L100-L106). Unfortunately, if a sub-expression changes the hash will not be updated :(

pocke:
True.
I've tried removing the cache, but it is slower than origin/master. So, it seems to be difficult to solve this problem....

PR #1067, opened 2017-06-30
Add plaintext fingerprint to warning hash

brentjo:
Occasionally when comparing two brakeman reports to find new warnings introduced by a code change, it will report an old warning and it's not immediately apparent why the code change caused the warning to reappear. You would assume it's because the fingerprint changed, but just seeing that two hashes are different doesn't give you any context for why it changed. This PR breaks up the one fingerprint method into two — one to generate the string and the other to hash it. This allows you to pass the plaintext fingerprint into the warning hash.

It's entirely possible this PR is just adding redundant data to the hash, and we would be fine using the data already in the hash (since it makes up most of the fingerprint), but having access to the fingerprint string would tell you with certainty why a fingerprint changed. I feel like people might find the fingerprint text useful to have, though perhaps it's not needed for most use cases.

### DIFF
--- a/lib/brakeman/warning.rb
+++ b/lib/brakeman/warning.rb
@@ -208,13 +208,16 @@ class Brakeman::Warning
    output
   end
 
-  def fingerprint
+  def fingerprint_as_s
     loc = self.location
     location_string = loc && loc.sort_by { |k, v| k.to_s }.inspect
     warning_code_string = sprintf("%03d", @warning_code)
     code_string = @code.inspect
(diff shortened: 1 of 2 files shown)


presidentbeef:
Hi Brent,

Thank you for the contribution.

However, I don't think adding this information to the JSON output is worth when I don't believe this is something very many users need very often.

What would be cool is if the `--compare` option could indicate why a fingerprint has changed...I'm not sure how hard that would be though.

presidentbeef:
Whoops.

Alternatively, maybe this could be optional information?

PR #1076, opened 2017-08-09
Modularize bin/brakeman

presidentbeef:
Move most of `bin/brakeman` into `lib/brakeman/commandline.rb`

* Simplify `bin/brakeman`
* Easier for people to extend if they want
* Possible to test now!


bin/brakeman
```diff
@@ -3,100 +3,6 @@
 $:.unshift "#{File.expand_path(File.dirname(__FILE__))}/../lib"
 
 require 'brakeman'
-require 'brakeman/options'
-require 'brakeman/version'
```
zimbatm: `require 'brakeman/version'` is missing now leading to a constant error

presidentbeef: Thanks, fixed with https://github.com/presidentbeef/brakeman/pull/1079

PR #1080, opened 2017-08-18
Delete unneeded test

gazayas:
I didn't write in an assertion here because the test name said "in_warnings"

I'm assuming the test wants to check how the relative path shows up when an error is thrown? If not, please let me know.

Thanks :octocat:

### DIFF
--- a/test/tests/brakeman.rb
+++ b/test/tests/brakeman.rb
@@ -26,16 +26,6 @@ class BrakemanTests < Minitest::Test
     assert_equal File.join(absolute_path, "Gemfile"), at.expand_path("Gemfile")
   end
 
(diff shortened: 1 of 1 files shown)


presidentbeef:
Hi Gabriel,

To be honest, I think that test was accidentally committed and doesn't make any sense. `AppTree` doesn't control the paths output with warnings.

Additionally, paths are tested here: https://github.com/presidentbeef/brakeman/blob/master/test/tests/json_output.rb#L44-L46

gazayas:
Hello Justin,

Thanks for the reply. I went ahead and deleted the whole tests, let me know what you think.

presidentbeef:
👍

PR #1091, opened 2017-09-04
Fix some issuse from 'codeclimate'

IgorPolyakov:
I wanted to correct some of the comments that generate "codeclimate"


presidentbeef:
Not a huge fan of the things rubocop worries about, but I appreciate the effort.

Issue #1100, opened 2017-09-25
Document --no-pager in 4.0.0 changes


bquorning:
Perhaps add a bit of documentation on the `--no-pager` option. When running Brakeman on CI, versions < 4 would exit with a status code, while version 4 will show its output in a pager (`less`?) and never exit.

The Pager was added in #1098.

bquorning:
Perhaps update https://github.com/presidentbeef/brakeman/blob/master/OPTIONS.md as well.

libo:
It looks like the breaking change is here:
https://github.com/presidentbeef/brakeman/blob/30c7e1cd108efeea2ee2a98f1af5d32954d182f8/lib/brakeman/options.rb#L243

And as @bquorning mention there is no mention of `--[no-]pager` in `OPTIONS.md` :-)

presidentbeef:
@bquorning what CI are you using? I want to make sure it's supported with some changes I'm making.

bquorning:
I’m using Travis.

presidentbeef:
Thanks @bquorning.

I've released 4.0.1 with #1101 to automatically disable the pager for most popular CIs and tested on Travis.
[OPTIONS.md](https://github.com/presidentbeef/brakeman/blob/master/OPTIONS.md) was updated by @nikolalsvk. The website has been updated as well: https://brakemanscanner.org/docs/options/

Of course the release post has this info, too, but you beat me to it: https://brakemanscanner.org/blog/2017/09/25/brakeman-4-dot-0-released/

daaain:
Just wanted to say I had this problem and adding `CI=true` to the command helped!

It took me a while to find the issue though, so leaving the error message here for search purposes: `press enter/return to continue or q to stop`.

PR #1101, opened 2017-09-25
Attempt to detect CI env and disable pager

presidentbeef:
Checks for `ENV["CI"]` which it appears most CIs are using.

Also fix output when `less` is not available or exceptions are raised.

Helps with #1100

### DIFF
--- a/lib/brakeman.rb
+++ b/lib/brakeman.rb
@@ -408,7 +408,11 @@ module Brakeman
   private_class_method :write_report_to_formats
 
   def self.page_output text
-    if system("which less")
+    ci = ENV["CI"]
+
+    if ci.is_a? String and ci.downcase == "true"
+      puts text
+    elsif system("which less")
       # Adapted from https://github.com/piotrmurach/tty-pager/
       write_io = open("|less -R", 'w')
       pid = write_io.pid
@@ -421,7 +425,7 @@ module Brakeman
       load_brakeman_dependency 'highline'
       h = ::HighLine.new
       h.page_at = :auto
-      h.say tracker.report.format(output_formats.first)
+      h.say text
     end
   rescue Errno::ECHILD
     # on jruby 9x waiting on pid raises (per tty-pager)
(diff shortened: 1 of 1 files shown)


lib/brakeman.rb
```diff
@@ -408,7 +408,11 @@ def self.write_report_to_formats tracker, output_formats
   private_class_method :write_report_to_formats
 
   def self.page_output text
-    if system("which less")
+    ci = ENV["CI"]
+
+    if ci.is_a? String and ci.downcase == "true"
```
grosser: should be `$stdout.tty?` instead ?

grosser:
not a fan of this option ... makes it behave strangely ... rubocop does not do that either ... I guess `git diff` does ... but for now always disabling it ...

lib/brakeman.rb
presidentbeef: See https://github.com/presidentbeef/brakeman/blob/ab3fc23bfdef7f7ea22ba2fee8641592793263cf/lib/brakeman.rb#L400

presidentbeef:
What is "behaving strangely"?

grosser:
Nice, that should solve it :)

idk how to describe it ... just feels weird that it's hanging after each
run ...

like git diff does not page when the output is short ... but brakeman still
pages even if there is less than a page of output ...
maybe just me being used to scroll up after each run ...

presidentbeef:
It takes some getting used to, but I think overall it's better now that the output tends to be longer.

There is an option for `less` to exit if the output is short. I can add that for the next release.

alex-hall:
Just got bit by this in our build and couldn't find anything in the docs related to the CI flag. 

Open for a pull request to add this information?

presidentbeef:
Well, the CI flag is set by CI providers like Travis, Circle, etc. It wasn't really intended to be more than that...

Is it more convenient to set an environment variable instead of just using the `--no-pager` option?

PR #1108, opened 2017-10-08
Update to use latest Code Climate test reporter client

noahd1:
* Using different matrix format so that Code Climate env var is only set for one ruby version
* Only running before and after script blocks if env var is set

Example: https://travis-ci.org/noahd1/brakeman/builds/285219207


presidentbeef:
:cool:

PR #1109, opened 2017-10-10
Support "app_path" configuration for Code Climate engine

noahd1:
This allows users of the Code Climate engine to specify a sub-directory to scan (something Brakeman itself has supported for a long time).

This support required 2 changes:

1. Scanning the subdirectory which is specified as "app_path" under the brakeman "config" block in a .codeclimate.yml
2. After scanning, ensuring that paths in reported issues are prefixed with the specified "app_path". This ensures that reported paths are from the root of the project (as required by Code Climate) instead of from the root of the application.


presidentbeef:
Seems okay to me.

PR #1126, opened 2017-11-30
Code Climate engine: fix `include_paths` when using custom app path

wfleming:
Brakeman (very reasonably) expects the `only_files` options to be
relative to the `app_path` option. The support that was added to the
Code Climate engine for a custom `app_path` option did not take
`include_paths` into account, so the normal result of using a custom
`app_path` would have ended up being that *nothing* would get analyzed:
because with a custom `app_path` of, say, `foo`, an include path might
include something like `foo/bar` when it needs to be adjusted to just
`bar`.

The fix here is to strip the `app_path` prefix from all matching
`include_paths`, and drop non-matching `include_paths` entirely.

### DIFF
--- a/lib/brakeman/codeclimate/engine_configuration.rb
+++ b/lib/brakeman/codeclimate/engine_configuration.rb
@@ -67,11 +67,26 @@ module Brakeman
       end
 
       def stripped_include_paths(prefix)
+        subprefixes = path_subprefixes(prefix)
         engine_config["include_paths"] && engine_config["include_paths"].map do |path|
-          if path && path.start_with?(prefix)
-            path.sub(%r{^#{prefix}/?}, "")
+          next unless path
+          if path.start_with?(prefix)
+            path.sub(%r{^#{prefix}/?}, "./")
+          elsif subprefixes.any? { |subprefix| path =~ %r{^#{subprefix}/?$} }
+            "./"
           end
         end.compact
+      end
+
+      def path_subprefixes(path)
+        Pathname.new(path).each_filename.inject([]) do |memo, piece|
+         memo <<
+           if memo.any?
+             File.join(memo.last, piece)
+           else
+             File.join(piece)
+           end
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


presidentbeef:
Looking at the tests...does this mean `include_paths` *must* include the `app_path` if both are specified?

wfleming:
User's don't specify `include_paths` explicitly: they specify `exclude_patterns`, and the Code Climate CLI effectively inverts those into the `include_paths` that an engine *can* report on.

So for the `include_paths` to not include the `app_path`, a user would have probably supplied a config like:

```
plugins:
  brakeman:
    enabled: true
    config:
      app_path: "foo"
exclude_patterns:
  - "foo/"
```

That's not really a sensible config, and in that case ending up with an empty `@configured_options[:only_files]` seems like the correct behavior to me.

You've made me realize there is, however, an edge case with my current implementation that would incorrectly set an empty `only_files`: if the `app_path` was something deeply nested like `foo/bar/baz`, and the include paths included `foo/` but hadn't expanded below that, this would incorrectly result in an empty file list right now, when it should actually be something like `["."]`. I'll try to fix that edge case today.

wfleming:
Thanks for noticing that edge case @presidentbeef, just pushed up a slightly tweaked approach.

presidentbeef:
Oh.

Then shouldn't this use `skip_files` instead? I don't like `only_files` in general, as Brakeman needs to see the whole application to be effective.

wfleming:
Users configure `exclude_patterns`, but they're not made available to 
engines, sorry. Engines have to use `include_paths` to determine what to 
analyze. We decided a long time ago to only make the include paths 
available to the engine since it limits the possibilities of inconsistent 
behavior from different engines/languages choosing to interpret glob 
patterns slightly differently.

presidentbeef:
If a pull request from Code Climate fails the Code Climate check...? :grimacing:

wfleming:
Heh, my bad, I didn't check locally before I pushed. Fixed.

presidentbeef:
Thanks!

wfleming:
Thank you!

PR #1134, opened 2017-12-05
Add SQL safe methods configurable option

abrom:
Adds `sql-safe-methods` option, similar to `safe-methods` and `url-safe-methods` to minimise false positives by allowing customisation of the list of 'safe' SQL injection methods.

The use case I have is for a common method used to pull in application configuration and wrap a column with a timezone offset. Trying to avoid wrapping the helper method with `sanitize_sql` or similar.

I see that you've had some reservations with these 'safe' method overrides in the past but they do appear to be here to stay? Thoughts?

### DIFF
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -211,6 +211,10 @@ To indicate certain methods return properly escaped output and should not be war
 
     brakeman --safe-methods benign_method_escapes_output,totally_safe_from_xss
 
+Brakeman warns about use of SQL injection for user methods. There are times when this is not applicable, such as methods where the result has been safely wrapped. Safe methods can be ignored with
+
+    brakeman --sql-safe-methods method_safe_from_sql_injection,another_safe_sql_method
+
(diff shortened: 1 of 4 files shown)


presidentbeef:
Hi Andrew,

I appreciate your work, but I'm still not a fan. For one, "safe" method options end up being misleading when they don't get applied everywhere people think they should.

Secondly, I am trying to avoid a proliferation of these options. The current ones were added before ignoring false positives was an option.

I would prefer to have people use the generic "ignore" functionality as needed.

abrom:
Sure, totally understand. I went with `safe` to keep it in line with the existing configuration options. Maybe `[url_]ignore_methods` would have been a better terminology for the functions? Point being somewhat moot..

The 'ignore' functionality does work just fine, albeit for my purpose a tad repetitive given it's essentially the same warning each time :)

On a side note, there would appear to be nothing stopping use of a function named 'sanitize' (or any of the other hard coded 'safe' methods) in the scope of the caller:

```
def foo
  Bar.select(sanitize)
end

def sanitize
  params[:unsafe]
end
```

Not suggesting I have a solution for such a problem given the dynamic nature of Ruby, just a thought. Somewhat trivial way that someone might unwittingly (or otherwise) hide a pretty clear SQL injection nasty

PR #1178, opened 2018-04-20
Code Climate: omit leading dot from `only_files`

toddmazierski:
When an `app_path` is specified in the Code Climate configuration, the relative paths that `Brakeman::Codeclimate::EngineConfiguration#active_include_paths` generates (added in presidentbeef/brakeman#1126) all start with a dot-slash (ex. `./app/`). This unfortunately causes a bug downstream in `Brakeman::AppTree#select_only_files` where _no_ paths actually match the `@only_files` regular expression.

A potential reason we missed this bug in the first place is the repository we were testing with used an `app_path` of `app`, which outwardly appears to mostly work.

/cc @wfleming

### DIFF
--- a/lib/brakeman/codeclimate/engine_configuration.rb
+++ b/lib/brakeman/codeclimate/engine_configuration.rb
@@ -87,9 +87,9 @@ module Brakeman
 
       def stripped_include_path(prefix, subprefixes, path)
         if path.start_with?(prefix)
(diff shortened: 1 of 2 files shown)


presidentbeef:
Ugh, `only_files` should go away...

presidentbeef:
Refreshed my memory with #1126. Still kind of wish it could go away :grinning:

PR #1214, opened 2018-05-22
Improve logic for handling conditionals within shell interpolation

### LINKED ISSUE #1205: False positive in command injection with conditional constant string interpolation
### Background

Brakeman version: 4.3.0
Rails version: 5.1.6
Ruby version: 2.5.1

#### False Positive

This appears to be a new issue introduced in Brakeman 4.3.0 (thanks for the new gem version, by the way—lots of exciting goodies for us!).

This code:

```ruby
def foo(use_bar)
  `echo #{"bar" if use_bar} baz`
  `echo #{use_bar ? "bar" : "baz"}`
end
```

produces command injection warnings when it did not in Brakeman 4.2.1. I believe this should be considered a false positive because the interpolation does not actually add the user-supplied value, but I'm not sure how easy that is for Brakeman to check.

Thoughts?

JacobEvelyn:
Fixes #1205

### DIFF
--- a/lib/brakeman/checks/check_execute.rb
+++ b/lib/brakeman/checks/check_execute.rb
@@ -143,7 +143,11 @@ class Brakeman::CheckExecute < Brakeman::BaseCheck
       next if SAFE_VALUES.include? e
       next if shell_escape? e
 
-      if node_type? e, :or, :evstr, :dstr
+      if node_type? e, :if
+        if res = dangerous?(e.values[1..-1])
+          return res
+        end
+      elsif node_type? e, :or, :evstr, :dstr
         if res = dangerous?(e)
           return res
         end
--- a/test/apps/rails5.2/lib/shell.rb
+++ b/test/apps/rails5.2/lib/shell.rb
@@ -31,4 +31,12 @@ class ShellStuff
   def system_as_target
     !system("echo #{foo}")
   end
+
+  def interpolated_conditional
+    `echo #{"foo" if foo} bar`
+  end
+
+  def interpolated_ternary
+    `echo #{foo ? "bar" : "baz"}`
+  end
 end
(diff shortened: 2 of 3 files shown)


JacobEvelyn:
I wasn't sure if this should also be done for SQL injection. Thoughts?

lib/brakeman/checks/check_execute.rb
```diff
@@ -143,7 +143,11 @@ def dangerous? exp
       next if SAFE_VALUES.include? e
       next if shell_escape? e
 
-      if node_type? e, :or, :evstr, :dstr
+      if node_type? e, :if
+        if res = dangerous?(e.values[1..-1])
```
presidentbeef: I would prefer

`dangerous?(e.then_clause) or dangerous(e.else_clause)`

JacobEvelyn: Ah neat, didn't know about those methods. Will change!

JacobEvelyn: Actually, this doesn't seem to work, I believe because `dangerous?` calls `each_sexp` on its argument and those clauses aren't "high-level" enough in the tree to have that work as we want. Feel free to fiddle around with the code in case there's a way to do it that I missed though!

JacobEvelyn:
I just pushed a new commit which adds positive tests for this.

presidentbeef:
> I wasn't sure if this should also be done for SQL injection. Thoughts?

It already is: https://github.com/presidentbeef/brakeman/blob/master/lib/brakeman/checks/check_sql.rb#L454

presidentbeef:
Thanks! :popcorn:

JacobEvelyn:
Ah great, I missed that when looking through `check_sql.rb`. Thanks!

PR #1217, opened 2018-05-23
Allow `symbolize_keys` to be called on params

### LINKED ISSUE #1206: False positive SQL injection when using `symbolize_keys` on `ActionController::Parameters.to_hash`
### Background

Brakeman version: 4.3.0
Rails version: 5.1.6
Ruby version: 2.5.1

#### False Positive

First of all, thanks so much for releasing 4.3.0! It's been awesome seeing all of the improvements.

It looks like I simplified our code a bit too much when creating the reproducible case for #1180. While we were calling `to_hash` on the `ActionController::Parameters` object, we also call `symbolize_keys` on _that_ because [Sequel](http://sequel.jeremyevans.net) requires keys to be symbols. Here's an example:

```ruby
def index
  # Using Sequel, though don't think it matters.
  people = Person.where(index_params).all
  render json: people
end

def index_params
  # The `to_hash` is necessary for Sequel's `where` method.
  params.permit(:name).to_hash.symbolize_keys
end
```

I believe this is a false positive, as `symbolize_keys` doesn't really change the semantics of the hash. The Brakeman error disappears if I remove the `.symbolize_keys`.

**EDIT:** I've extracted part of this issue into a new issue: #1216.

JacobEvelyn:
Fixes #1206

### DIFF
--- a/lib/brakeman/checks/check_sql.rb
+++ b/lib/brakeman/checks/check_sql.rb
@@ -290,7 +290,7 @@ class Brakeman::CheckSQL < Brakeman::BaseCheck
     end
 
     if request_value? arg
-      unless call? arg and params? arg.target and [:permit, :slice, :to_h, :to_hash].include? arg.method
+      unless call? arg and params? arg.target and [:permit, :slice, :to_h, :to_hash, :symbolize_keys].include? arg.method
         # Model.where(params[:where])
         arg
       end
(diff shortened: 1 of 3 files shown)


presidentbeef:
:cool: Thanks!

JacobEvelyn:
Thanks for the quick merging of these two PRs, @presidentbeef! Selfishly, it would help us a lot if you could release a new gem version soon, but I'll let you determine when it makes the most sense for you. 😄

presidentbeef:
A release takes about a full day of work, so hard to say. Maybe very soon, schedule permitting.

That being said, master is pretty much always stable.

JacobEvelyn:
Gotcha, thanks for the info!

PR #1244, opened 2018-07-18
Ignore Tempfiles in FileAccess warnings

### LINKED ISSUE #1243: Brakeman warns about FileAccess param insertion for Tempfiles
### Background

Brakeman version: 4.3.1
Rails version: 5.1
Ruby version: 2.5

#### False Positive

*Full* warning from Brakeman:

```
{:warning_type=>"File Access",
 :warning_code=>16,
 :fingerprint=>
  "91c27983715e71528e8ad0c0fb1763d3df7bd79111369ff2c81c10ebf64e825c",
 :check_name=>"SendFile",
 :message=>
  "Parameter value used in file name near line 3: send_file(Tempfile.new(params[:file_name], \".txt\"))",
 :file=>
  "/Users/ckoller/panorama/brakeman/test/apps/rails5/app/controllers/file_controller.rb",
 :line=>3,
 :link=>"https://brakemanscanner.org/docs/warning_types/file_access/",
 :code=>"send_file(Tempfile.new(params[:file_name], \".txt\"))",
 :render_path=>nil,
 :location=>
  {:type=>:method,
   :class=>:FileController,
   :method=>:download_tempfile_with_params},
 :user_input=>"params[:file_name]",
 :confidence=>"Weak"}
```

Relevant code:

```ruby
class FileController < ApplicationController
  def download_tempfile_with_params
    send_file Tempfile.new(params[:file_name], ".txt")
  end
end
```

When Tempfiles are initialized with an argument for filename, the resulting filename has a set of unique characters appended to the end. Tempfile guarantees that the file it creates cannot already exist in the file system, using the modes `File::CREAT` and `File::EXCL` (see [Tempfile#initialize source code here](https://ruby-doc.org/stdlib-2.5.1/libdoc/tempfile/rdoc/Tempfile.html#method-c-new)).  This means that regardless of what the user chooses to enter as a filename, the user cannot maliciously access any other files on the system. 

```
Tempfile.new(["secrets", ".txt"])
=> #<Tempfile:/var/folders/cn/gbgnkncj6d9dl5f330zgp3ww0000gp/T/secrets20180718-30281-k5dut2.txt>
```

I'm working on a PR for this right now, and should have it up soon!

cmkoller:
Fixes #1243


presidentbeef:
Hi Christina,

Thank you for the PR! Please acknowledge that copyright for all contributions will be assigned to Synopsys, Inc. per our [contribution guidelines](https://github.com/presidentbeef/brakeman/blob/master/.github/contributing.md).

The approach seems a little bit broad - what about just checking the `file_name`? Preferably something like `call? file_name and file_name.target == s(:const, :Tempfile)`.

presidentbeef:
Right now you are checking the entire call, but just a couple lines down there's the `file_name` (which is just the first argument to the call). So you can check that instead. I assume we want to ignore any calls on `Tempfile`, so checking that `file_name` is a `call` expression and checking that the target is the `Tempfile` constant should be enough. That's the code I showed above.

There are a number of convenience methods like `call?` in [Util](https://github.com/presidentbeef/brakeman/blob/master/lib/brakeman/util.rb#L146) and some convenience methods on `Sexp` in [here](https://github.com/presidentbeef/brakeman/blob/master/lib/ruby_parser/bm_sexp.rb#L132).

Let me know if that doesn't make sense.

cmkoller:
@presidentbeef Thanks for clarifying! I pushed an update using your suggestion :) let me know what you think!

JacobEvelyn:
Just wondering if there's any reason for the delay here? This change would be really nice!

PR #1254, opened 2018-08-29
Fix an issue that fails if secrets.yml is an empty file.

naokikimura:
If `config/secrets.yml` is an empty file, the following error occurs:

```
== Errors ==

Error: undefined method `[]' for false:FalseClass
Location: /usr/local/opt/asdf/installs/ruby/2.3.7/lib/ruby/gems/2.3.0/gems/brakeman-4.3.1/lib/brakeman/checks/check_session_settings.rb:127:in `check_secrets_yaml'

```


naokikimura:
Hi @presidentbeef. 

Added test covering the scenario.

presidentbeef:
Hi @naokikimura, thank you for this contribution. A full new test app was a bit much just to test an empty secrets.yml (sorry for the confusion). I added an empty `secrets.yml` to the Rails 5.2 test app and merged your fix.

Thank you again!

PR #1259, opened 2018-09-11
Add more structured message construction

presidentbeef:
Updates to make warning message format more flexible and consistent.

Instead of creating a string for a warning message, it is instead a `Brakeman::Messages::Message` which is essentially a list of `Brakeman::Messages` types.

For example:

```ruby
 "Unescaped #{friendly_type_of @matched} in content_tag"
```
becomes
```ruby
 msg("Unescaped ", msg_input(@matched), " in ", msg_code("content_tag"))
```

The output value becomes something like
```ruby
Message(Plain("Unescaped "), Input(@matched), Plain(" in "), Code("content_tag"))
```

This way values like code, versions, input types, etc. can be formatted at output time depending on the report format (e.g. HTML). In the future this may help with translations.


presidentbeef:
I'd better document more of the changes:

- Messages are no longer truncated in HTML reports. But code at the end of the message might still be  truncated if there is a huge chunk of code being displayed.
- CVEs in HTML reports are links to MITRE
- Links in HTML reports now correctly use `rel="noreferrer"`

PR #1270, opened 2018-10-08
Do not warn about *_url on models in redirects

MC-Squared:
## Issue
I have an app that uses draper decorators, one of which provides a `step_url` method, which produces warnings. Changing it to `step_path` removes the warnings, but is slightly misleading because it uses the `url_for` helper.

## Fix
This is basically just an add-on to #1164, so that model methods ending in `_url` also do not generate warnings.

### DIFF
--- a/lib/brakeman/checks/check_redirect.rb
+++ b/lib/brakeman/checks/check_redirect.rb
@@ -79,7 +79,7 @@ class Brakeman::CheckRedirect < Brakeman::BaseCheck
     end
 
     if res = has_immediate_model?(arg)
-      unless call? arg and arg.method.to_s =~ /_path/
+      unless call? arg and arg.method.to_s =~ /_path|_url/
         return Match.new(immediate, res)
       end
     elsif call? arg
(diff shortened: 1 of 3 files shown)


presidentbeef:
Hi Jonathan,
:grey_question: 
Thank you for your contribution.

However, this seems to contradict the intention of this check. `*_path` methods are ignored because they typically do not return full URLs, so an attacker cannot redirect to a malicious site.

`*_url` methods are much more likely to contain full URLs, possibly controller by an attacker. That is exactly what this check is trying to warn about.

:thinking:

MC-Squared:
Hi Justin,

I agree a `_url` method is more likely to contain a full path. My understanding (perhaps incorrect) was that the check was more about using user input - i.e. via `params`.

I'm happy to have this PR rejected and will update our codebase instead if you think that's is the better approach to avoid potential false negatives.

presidentbeef:
Yes, I do not think this change makes sense at this time.

Thanks!

Issue #1284, opened 2018-12-04
Slim v4.0 parser failure


Jakenberg:
### Background

Brakeman version: 4.3.1
Rails version: 5.0.2
Ruby version: 2.3.7

Link to Rails application code: N/A

#### Parse Error

The following error is produced by running `brakeman -d` in project directory:
```
== Errors ==

Error: Expected tag   [project_folder]/app/views/layouts/application.html.slim, Line 16, Column 62     javascript(nonce==content_security_policy_script_nonce):                                                             ^
Location: [project_folder]/app/views/layouts/application.html.slim:16
[home_folder]/.rvm/gems/ruby-2.3.7/gems/brakeman-4.3.1/bundle/ruby/2.5.0/gems/slim-3.0.7/lib/slim/parser.rb:513:in `syntax_error!'
[home_folder]/.rvm/gems/ruby-2.3.7/gems/brakeman-4.3.1/bundle/ruby/2.5.0/gems/slim-3.0.7/lib/slim/parser.rb:367:in `parse_tag'
[home_folder]/.rvm/gems/ruby-2.3.7/gems/brakeman-4.3.1/bundle/ruby/2.5.0/gems/slim-3.0.7/lib/slim/parser.rb:248:in `parse_line_indicators'
[home_folder]/.rvm/gems/ruby-2.3.7/gems/brakeman-4.3.1/bundle/ruby/2.5.0/gems/slim-3.0.7/lib/slim/parser.rb:192:in `parse_line'
[home_folder]/.rvm/gems/ruby-2.3.7/gems/brakeman-4.3.1/bundle/ruby/2.5.0/gems/slim-3.0.7/lib/slim/parser.rb:95:in `call'
[home_folder]/.rvm/gems/ruby-2.3.7/gems/brakeman-4.3.1/bundle/ruby/2.5.0/gems/temple-0.7.7/lib/temple/engine.rb:50:in `block in call'
[home_folder]/.rvm/gems/ruby-2.3.7/gems/brakeman-4.3.1/bundle/ruby/2.5.0/gems/temple-0.7.7/lib/temple/engine.rb:50:in `each'
[home_folder]/.rvm/gems/ruby-2.3.7/gems/brakeman-4.3.1/bundle/ruby/2.5.0/gems/temple-0.7.7/lib/temple/engine.rb:50:in `inject'
[...]

PR #1285, opened 2018-12-04
upgrade version upper boundary of slim to v4.0.1

Jakenberg:
This is in regards to issue #1284.

### DIFF
--- a/gem_common.rb
+++ b/gem_common.rb
@@ -17,7 +17,7 @@ module Brakeman
       spec.add_dependency "erubis", "~>2.6"
       spec.add_dependency "haml", ">=3.0", "<5.0"
       spec.add_dependency "sass", "~>3.0", "<3.5.0"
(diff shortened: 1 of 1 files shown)


presidentbeef:
@Jakenberg Thanks for looking into this.

Can you add an example of the new syntax to a Slim template in https://github.com/presidentbeef/brakeman/tree/master/test/apps/rails5.2 for testing? Thanks!

Jakenberg:
No problem. Added what I think you mean. Let me know if it should be different 👍

presidentbeef:
Okay, asking you to add the test was a mistake since Ruby 1.9.3 is not supported by Slim 4.0 :(

Do you mind backing that change out and then I'll merge? Thanks again!

Jakenberg:
Bummer :( Updated

presidentbeef:
Thanks!

PR #1286, opened 2018-12-04
Upgrade RubyParser to 3.12

presidentbeef:
Also, minor versions are not as scary now (especially with the dependency vendoring), so be more liberal there.


djpate:
:+1: I need this as well

presidentbeef:
Just a note that there is a (probably rare) regression introduced by 3.12.0: https://github.com/seattlerb/ruby_parser/pull/272

PR #1287, opened 2018-12-05
Upgrade ruby_parser to 3.12.0

djpate:
In order to allow for rescue in block syntax, we need to bump to ruby_parser `3.12.0`.

Cutting a release would be cool too!


djpate:
nvm just saw the other PR

PR #1311, opened 2019-02-06
Relax Sass dependency to ~>3.0

### LINKED ISSUE #1295: Please consider relaxing sass dependency
Hello. The dependency on `sass` was last set to <3.5.0 over a year ago, and released in version 4.3.1.  This is preventing us from upgrading brakeman because we use `scss_lint`, which depends on sass version `~> 3.5.5`. From the commit message, it looks like the reason for the version constraint was to support Ruby 1.9. Would it be possible to add some logic so that folks using the latest version of Ruby can use the latest version of Sass? 

Perhaps would it make sense to stop supporting 1.9? Support for 1.9.3 ended in February 2015: https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/

presidentbeef:
Closes #1295

Note that this means Ruby 1.9 will no longer be supported in the `brakeman` gem release.

### DIFF
--- a/gem_common.rb
+++ b/gem_common.rb
@@ -17,7 +17,7 @@ module Brakeman
       spec.add_dependency "highline", ">=1.6.20", "<2.0"
       spec.add_dependency "erubis", "~>2.6"
       spec.add_dependency "haml", ">=3.0", "<5.0"
-      spec.add_dependency "sass", "~>3.0", "<3.5.0"
+      spec.add_dependency "sass", "~>3.0"
       spec.add_dependency "slim", ">=1.3.6", "<=4.0.1"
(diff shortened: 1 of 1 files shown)


monfresh:
Thanks!

presidentbeef:
Newer versions pull in `sass-listen` which pulls in `ffi` (requires build tools), `rb-fsevent`, and `rb-inotify`. Kind of annoying...also `sass` is almost EOL.

My current thought is to drop the `sass` dependency and have a different solution for handling filters in general.

presidentbeef:
Follow up is in #1314

PR #1332, opened 2019-03-18
Fix README for Docker instructions

JacobEvelyn:
I don't know a ton about Docker so I might just be missing something but I had to make this tweak to get `docker run` to work.


presidentbeef:
You are right - I had an old image lying around named "brakeman" which threw me off.

Thanks!

PR #1347, opened 2019-04-18
Use relative paths for __FILE__

presidentbeef:
RubyParser converts `__FILE__` to the actual file name at parse time. Brakeman was passing in the full absolute path to use for that file name. That would cause fingerprints of warnings involving `__FILE__` to change based on where the code happened to be when it was scanned.

Instead, changed to passing in a relative file path for better fingerprint stability.

This also affects parse error messages.

### DIFF
--- a/test/tests/rails52.rb
+++ b/test/tests/rails52.rb
@@ -13,7 +13,7 @@ class Rails52Tests < Minitest::Test
       :controller => 0,
       :model => 0,
       :template => 5,
-      :generic => 16
+      :generic => 17
     }
   end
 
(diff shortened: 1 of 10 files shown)


test/tests/rails52.rb
```diff
@@ -13,7 +13,7 @@ def expected
       :controller => 0,
       :model => 0,
       :template => 5,
-      :generic => 16
+      :generic => 17
```
qsphan: Why is it changed to 17?

presidentbeef: These numbers indicate how many issues are in the test suite. Since this change added a test case, the expected number of issues increased.

PR #1360, opened 2019-06-05
Avoid assigning nil line numbers to Sexps

presidentbeef:
Since that raises an error in latest SexpProcessor.

### DIFF
--- a/lib/brakeman/processors/controller_processor.rb
+++ b/lib/brakeman/processors/controller_processor.rb
@@ -192,8 +192,8 @@ class Brakeman::ControllerProcessor < Brakeman::BaseProcessor
 
     filter_name = ("fake_filter" + rand.to_s[/\d+$/]).to_sym
     args = exp.block_call.arglist
-    args.insert(1, Sexp.new(:lit, filter_name))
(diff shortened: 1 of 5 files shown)


presidentbeef:
Test coverage coming in another PR, because problems were uncovered 😂

lib/brakeman/processors/controller_processor.rb
```diff
@@ -192,8 +192,8 @@ def add_fake_filter exp
 
     filter_name = ("fake_filter" + rand.to_s[/\d+$/]).to_sym
     args = exp.block_call.arglist
-    args.insert(1, Sexp.new(:lit, filter_name))
-    before_filter_call = make_call(nil, :before_filter, args)
+    args.insert(1, Sexp.new(:lit, filter_name).line(exp.line))
+    before_filter_call = make_call(nil, :before_filter, args).line(exp.line)
```
zenspider: These failed on my non-nil change?

presidentbeef: This may have been general cleanup to ensure Sexps have line numbers to prevent problems later.

PR #1367, opened 2019-06-19
Add reverse tabnabbing check

linosgian:
## Description
[Anchor HTML tags](https://en.ryte.com/wiki/Anchor_Tag) have an attribute that describes how the linked document should be opened, named `target`. Some of the accepted values are the following:
- `_blank` tells the browser to open the linked document in a new tab/window
- `_self` (the default value) opens the document on the same tab.

[Reverse tabnabbing](https://www.owasp.org/index.php/Reverse_Tabnabbing) is a security vulnerability where if an adversary can create anchor tags with `target: _blank`,  they can redirect the victim to any domain they want. This is possible because the newly opened tab is able to communication with the parent tab through `window.opener`. The most common exploitation route here is to redirect the victim to a phishing using the following PoC:
` window.opener.location = "https://attacker.co"; `.

Let's consider the following scenario:
We have a forum-type of website where users' comments can contain URLs. These URLs get automatically converted to `<a>` tags with the `target: _blank` element so that users original tab stays intact (clicked document opens in a new tab). In this case, a malicious user can submit a link that redirects the victim to their own malicious website and present them with our forum's (fake) login page. The user submits their credentials to the adversary's website.

### Fix
The fix for this is quite simple: add the `noopener` value to the `rel` attribute as such:
`<a href="//attacker.co" target="_blank" rel="noopener">`

## Proposed check
This PR adds a new _**optional**_ check that looks for such cases. More specifically, we look for `link_to` calls where `target: "_blank"` is passed to `html_opts` and either:

- the `rel` attribute is not supplied at all
- the `noopener` is not supplied.

I've tried to follow the projects conventions but there will be room for improvement.

Awaiting your feedback! :)

### DIFF
--- a/lib/brakeman/checks/check_reverse_tabnabbing.rb
+++ b/lib/brakeman/checks/check_reverse_tabnabbing.rb
@@ -1,0 +1,43 @@
+require 'brakeman/checks/base_check'
+
+class Brakeman::CheckReverseTabnabbing < Brakeman::BaseCheck
+  Brakeman::Checks.add_optional self
+
+  @description = "Checks for reverse tabnabbing cases on 'link_to' calls"
+
+  def run_check
+    calls = tracker.find_call :methods => :link_to
+    calls.each do |call|
+      process_result call
+    end
+  end
+
+  def process_result result
+    return unless original? result and result[:call].third_arg
+
+    html_opts = result[:call].third_arg
+    _, target = hash_access html_opts, :target
+    _, rel = hash_access html_opts, :rel
+    return unless target
+
+    target_url = result[:call].second_arg
+
+    # `_url` and `_path` lead to urls on to the same origin.
+    # That means that an adversary would need to run javascript on
+    # the victim application's domain. If that is the case, the adversary
+    # already has the ability to redirect the victim user anywhere.
+    if target_url[0] == :call then
+      func_s = target_url.method.to_s
+      return unless !func_s.end_with?("url") && !func_s.end_with?("path")
+    end
+
+    if target == "_blank" && (!rel || (rel && !rel.include?("noopener"))) then
+      warn :result => result,
+        :warning_type => "Reverse Tabnabbing",
+        :warning_code => :reverse_tabnabbing,
+        :message => "The newly opened tab can control the parent tab's " +
+                    "location, thus redirect it to a phishing page",
+        :confidence => :weak
+    end
+  end
+end
--- a/lib/brakeman/checks/check_reverse_tabnabbing.rb
+++ b/lib/brakeman/checks/check_reverse_tabnabbing.rb
@@ -23,25 +23,16 @@ class Brakeman::CheckReverseTabnabbing < Brakeman::BaseCheck
 
     target_url = result[:call].second_arg
 
-    # `_url` and `_path` lead to urls on to the same origin.
-    # That means that an adversary would need to run javascript on
-    # the victim application's domain. If that is the case, the adversary
-    # already has the ability to redirect the victim user anywhere.
-    if call? target_url then
-      func_s = target_url.method.to_s
-      return unless !func_s.end_with?("url") && !func_s.end_with?("path")
-    end
-
     rel = hash_access html_opts, :rel
+    confidence = :medium
 
-    if !rel then
-      confidence = :medium
-    elsif !string?(rel)
-      return
-    elsif rel.include?("noopener") && !rel.include?("noreferrer") then
-      confidence = :weak
-    else
-      confidence = :medium
+    if rel && string?(rel) then
+      rel = rel.value
+      return if rel.include?("noopener") && rel.include?("noreferrer")
+
+      if rel.include?("noopener") ^ rel.include?("noreferrer") then
+        confidence = :weak
+      end
     end
 
     warn :result => result,
--- a/lib/brakeman/checks/check_reverse_tabnabbing.rb
+++ b/lib/brakeman/checks/check_reverse_tabnabbing.rb
@@ -21,7 +21,15 @@ class Brakeman::CheckReverseTabnabbing < Brakeman::BaseCheck
     target = hash_access html_opts, :target
     return unless target && string?(target) && target.value == "_blank"
 
-    target_url = result[:call].second_arg
+    target_url = result[:block] ? result[:call].first_arg : result[:call].second_arg
+
+    # `url_for` and `_path` calls lead to urls on to the same origin.
+    # That means that an adversary would need to run javascript on
+    # the victim application's domain. If that is the case, the adversary
+    # already has the ability to redirect the victim user anywhere.
+    # Also statically provided URLs (interpolated or otherwise) are also
+    # ignored as they produce many false positives.
+    return if !call?(target_url) || target_url.method.match(/^url_for$|_path$/)
 
     rel = hash_access html_opts, :rel
     confidence = :medium
--- a/lib/brakeman/checks/check_reverse_tabnabbing.rb
+++ b/lib/brakeman/checks/check_reverse_tabnabbing.rb
@@ -13,12 +13,13 @@ class Brakeman::CheckReverseTabnabbing < Brakeman::BaseCheck
   end
 
   def process_result result
-    return unless original? result and result[:call].third_arg
+    return unless original? result and result[:call].last_arg
 
-    html_opts = result[:call].third_arg
-    _, target = hash_access html_opts, :target
-    _, rel = hash_access html_opts, :rel
-    return unless target
+    html_opts = result[:call].last_arg
+    return unless hash? html_opts
+
+    target = hash_access html_opts, :target
+    return unless target && string?(target) && target.value == "_blank"
 
     target_url = result[:call].second_arg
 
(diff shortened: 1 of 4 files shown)


presidentbeef:
Okay, weird...I literally worked on this same check yesterday.

lib/brakeman/checks/check_reverse_tabnabbing.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def run_check
+    calls = tracker.find_call :methods => :link_to
+    calls.each do |call|
+      process_result call
+    end
+  end
+
+  def process_result result
+    return unless original? result and result[:call].third_arg
```
presidentbeef: This assumes only one form of `link_to`. There are several: https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to

You can check `result[:block]` to see if it's a block form and `result[:call].last_arg` can be used to get the options.

lib/brakeman/checks/check_reverse_tabnabbing.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    html_opts = result[:call].third_arg
+    _, target = hash_access html_opts, :target
```
presidentbeef: `hash_access` does not check if the value is actually a Hash node. It should be checked with `hash?` before calling `hash_access`.

There is no guarantee the value will be a string Sexp. The value should be checked with `string?` and then you can get the string value with `target.value`. Same for `rel` below.

lib/brakeman/checks/check_reverse_tabnabbing.rb
```diff
@@ -0,0 +1,43 @@
...
+    _, rel = hash_access html_opts, :rel
+    return unless target
+
+    target_url = result[:call].second_arg
+
+    # `_url` and `_path` lead to urls on to the same origin.
+    # That means that an adversary would need to run javascript on
+    # the victim application's domain. If that is the case, the adversary
+    # already has the ability to redirect the victim user anywhere.
+    if target_url[0] == :call then
```
presidentbeef: Use `call?` instead of accessing and checking the node type directly.

lib/brakeman/checks/check_reverse_tabnabbing.rb
```diff
@@ -0,0 +1,43 @@
...
+      func_s = target_url.method.to_s
+      return unless !func_s.end_with?("url") && !func_s.end_with?("path")
```
presidentbeef: Not sure I agree with checking for `url`. That could easily be an attacker-controllable value on a model.

In my version I did `url.method.match?(/^url_for$|_path$/)` (although I haven't checked for false positives yet).

lib/brakeman/checks/check_reverse_tabnabbing.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    if target == "_blank" && (!rel || (rel && !rel.include?("noopener"))) then
```
presidentbeef: `target` value can be checked up above where it is checked for `nil`.

From my research, you have to set both `noopener` and `noreferrer` because IE/Edge don't support `noopener` 😞 So my approach was weak confidence if `noopener` was used, medium confidence if neither were used.

test/tests/rails5.rb
```diff
@@ -747,4 +747,37 @@ def test_unscoped_find
       :code => s(:call, s(:const, :User), :find, s(:call, s(:params), :[], s(:lit, :id))),
       :user_input => s(:call, s(:params), :[], s(:lit, :id))
   end
+
+  def test_reverse_tabnabbing
```
presidentbeef: Duplicate code.

lib/brakeman/checks/check_reverse_tabnabbing.rb
```diff
@@ -0,0 +1,43 @@
+require 'brakeman/checks/base_check'
+
+class Brakeman::CheckReverseTabnabbing < Brakeman::BaseCheck
+  Brakeman::Checks.add_optional self
```
presidentbeef: I'm not sure it needs to be an optional. Depends on false positive rate, which I will look into.

lib/brakeman/checks/check_reverse_tabnabbing.rb
presidentbeef: Actually, here are my test cases:

```erb
<%= link_to '', 'something_static', target: '_blank' %> No warning
<%= link_to '', some_url, target: '_blank' %> Warn
<%= link_to some_url, target: '_blank' do -%>
  Warn
<% end %>
<%= link_to '', some_url, target: '_blank', rel: 'noopener' %> Weak warning
<%= link_to '', some_url, target: '_blank', rel: 'noreferrer' %> Weak warning
<%= link_to '', some_url, target: '_blank', rel: 'noopener noreferrer' %> No warning
<%= link_to some_url, target: '_blank', rel: 'noopener noreferrer' do -%>
  No Warning
<% end %>
```

lib/brakeman/checks/check_reverse_tabnabbing.rb
linosgian: Yes that might have slipped through,IE/Edge should (unfortunately) be covered. More info on this argument [here](https://caniuse.com/#feat=rel-noopener).

I'll push a fixup for the first part of the comment

lib/brakeman/checks/check_reverse_tabnabbing.rb
linosgian: You are right, and I've also tested this in conjuction with [this](https://github.com/rails/rails/issues/29893). And links that use the `_url` helper are vulnerable. If the the `_url` helper method is used (the same is true for `url_for` calls), the `X-Forwarded-Host` header can be used to poison the cache and so control the domain of the resulting `<a>` tags for as long as the cache holds the cache entry.

So you are specifically looking for `url_for`/`_path` or you are ruling those out?

linosgian:
I've pushed a fixup after going through most of your comments, i'll work on the rest later on.

lib/brakeman/checks/check_reverse_tabnabbing.rb
linosgian: I am not sure ruling out static strings as false positives would be the right choice here. A statically provided URL (e.g. https://example.com) still leaves the victim vulnerable to that specific website. This might or might not be exploitable, depending on the security of the referred website. We essentially trust that this website won't have a security issue that let's the attacker add javascript to the resource we provide a link to, on our application.

I think that these cases have to be included in the results.

Also, I've tested the case for `noreferrer` only and both Chrome and Firefox seem to forbid the access to `window.opener.location` from the new tab. Thus the issue is effectively solved. This seems weird, since `noreferrer` is supposed to only drop the `Referrer:` header when issuing a request to the newly opened tab.

What do you think about the first case?

lib/brakeman/checks/check_reverse_tabnabbing.rb
linosgian: Alright. I am not sure there is a legitimate use of the `window.opener` communication, a case where a user actually wants the newly opened tab to affect the parent tab.. I think that a suggestion to add the `noopener` everywhere wouldn't hurt, and in fact [improve performance](https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/). 

The `noreferrer` might have negative side-effects in cases where the developer actually wants the `Referer` header to be included in the request sent to the new tab, for analytics and so on..

lib/brakeman/checks/check_reverse_tabnabbing.rb
presidentbeef: I think `url_for`/`_path` should be ignored. The issue with `X-Forwarded-Host` aside, these are most often going to be used for internal links.

lib/brakeman/checks/check_reverse_tabnabbing.rb
presidentbeef: > We essentially trust that this website won't have a security issue that let's the attacker add javascript to the resource we provide a link to, on our application.

I hear you and this is a valid concern, but one that is a bit below the threshold for many organizations to worry about.

> I think that these cases have to be included in the results.

However, I disagree. It creates too many false positives. I see them in my test corpus.

Examples:

https://github.com/18F/myusa/blob/devel/app/views/devise/sessions/new.html.erb#L9
https://github.com/alphagov/whitehall/blob/master/app/views/admin/edition_workflow/confirm_unpublish.html.erb#L16
https://github.com/amahi/platform/blob/master/app/views/layouts/_footer.html.slim#L6-L25

test/apps/rails5/app/views/users/show.html.erb
```diff
@@ -7,3 +7,16 @@
...
+<%= link_to some_url, target: '_blank' do -%>
+  Warn
+<% end %>
+<%= link_to '', some_url, target: '_blank', rel: 'noopener' %> Weak warning
+<%= link_to '', some_url, target: '_blank', rel: 'noreferrer' %> Weak warning
+<%= link_to '', some_url, target: '_blank', rel: 'noopener noreferrer' %> No warning
+<%= link_to some_url, target: '_blank', rel: 'noopener noreferrer' do -%>
+  No Warning
+<% end %>
+<%= link_to '', target: '_blank' %> No warning
```
linosgian: The URL could be absent entirely. This covers that case

lib/brakeman/checks/check_reverse_tabnabbing.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    target_url = result[:block] ? result[:call].first_arg : result[:call].second_arg
+
+    # `url_for` and `_path` calls lead to urls on to the same origin.
+    # That means that an adversary would need to run javascript on
+    # the victim application's domain. If that is the case, the adversary
+    # already has the ability to redirect the victim user anywhere.
+    # Also statically provided URLs (interpolated or otherwise) are also
+    # ignored as they produce many false positives.
+    return if !call?(target_url) || target_url.method.match(/^url_for$|_path$/)
```
linosgian: I figured that variables are represented by :calls as they could be no-argument function calls.
I think the !call?(target_url) covers all the cases we need

presidentbeef:
Sorry for the delay!

I did some squashing and adjusting, which ended up making it look like this was not merged. But it is! :sparkle: 

Thank you for your contribution and patience!

linosgian:
No worries, cheers! :+1:

PR #1375, opened 2019-07-08
Allow file access with ActiveStorage::Filename#sanitized

### LINKED ISSUE #1348: ActiveStorage::Filename#sanitized false positive for File Access category
### Background

Brakeman version: 4.4.0
Rails version: 2.3.18
Ruby version: 2.3.7

#### False Positive

```
Confidence: Weak
Category: File Access
Check: SendFile
Message: Parameter value used in file name
Code: send_file("path/to/csv/#{ActiveStorage::Filename.new(params[:test]).sanitized}.csv", :disposition => "attachment", :encoding => "utf8", :filename => ("#{ActiveStorage::Filename.new(params[:test]).sanitized}.csv"))
```

Relevant code:

`ActiveStorage::Filename.new(params[:test]).sanitized`

I copied `ActiveStorage::Filename` implementation to prevent `File Access` attacks (https://edgeapi.rubyonrails.org/classes/ActiveStorage/Filename.html#method-i-sanitized), I was expecting that this change was going to fix all File Access brakeman errors, but they are still there.

Not sure if this is something because I'm not running Rails 5 or if it is something that we can change in Brakeman to not report File Access errors when `sanitized` method is used.

Thanks for the great gem 🙌

tejasbubane:
Closes #1348

### DIFF
--- a/docs/warning_types/file_access/index.markdown
+++ b/docs/warning_types/file_access/index.markdown
@@ -9,3 +9,9 @@ will raise an error like
     Cookie value used in file name near line 4: File.open("/tmp/#{cookie[:file]}")
 
 This type of vulnerability can be used to access arbitrary files on a server (including `/etc/passwd`.
+
+If you are using `ActiveStorage`, use [sanitized](https://api.rubyonrails.org/classes/ActiveStorage/Filename.html#method-i-sanitized) URLs:
+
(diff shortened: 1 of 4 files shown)


presidentbeef:
Seems to handle a limited case, but that's okay for now.

PR #1378, opened 2019-07-10
Document known issue: multiple gem requirements

jaredbeck:
It's fairly rare for people to use multiple `gem` requirements; more common in gemspecs than in app `Gemfile`s. But, still I recommend we recognize this issue by adding this comment.


presidentbeef:
I agree at some point the version checking can probably switch to using classes from RubyGems.

PR #1388, opened 2019-07-24
Fix ReverseTabnabbing warning message

stsc3000:
Hi!
The warning message for `ReverseTabnabbing` reads
> When opening a link in a new tab without setting `rel: "noopener noreferr"`, the new tab can control the parent tab's location. For example, an attacker could redirect to a phishing page.

when it should actually include `noreferrer` instead. The checks work as expected.
I'm not sure whether the tests need to be adapted for such a minor change.


presidentbeef:
Thank you! Nice to know people are paying attention 😊

PR #1389, opened 2019-07-24
Fix typo in example code in message for reverse tabnabbing

joergschiller:
Hi!

When brakemen detects Reverse Tabnabbing there is a type in the example code (`noreferr`).

```
Confidence: Medium
Category: Reverse Tabnabbing
Check: ReverseTabnabbing
Message: When opening a link in a new tab without setting `rel: "noopener noreferr"`, the new tab can control the parent tab's location. For example, an attacker could redirect to a phishing page.
```


presidentbeef:
Thank you! However, someone got there a few hours before you in #1388 😄

joergschiller:
Thank you! Shame on my, should have at least searched for that change in the existing PRs as well.

PR #1391, opened 2019-08-08
Fix exception in CheckCookieSerialization

philipturnbull:
The `and`/`or` precedence rules cause us to execute `setting.value == :hybrid`
even when `setting` is not a symbol, which raises an exception:

```
Sexp#value called on multi-item Sexp: `s(:colon2, s(:const, :Custom), :Serializer)`
```


presidentbeef:
Thanks @philipturnbull! How embarrassing :sweat_smile:

PR #1395, opened 2019-08-27
Add check for ERB.new that includes user input

mhickman:
This is pretty much just as bad as eval

example:
```ruby
bad_input = "<% User.delete_all %>"
ERB.new(bad_input).result
```

Adding a check to check_evaluation to also check for dangerous calls to ERB.new

### DIFF
--- a/docs/warning_types/template_injection/index.markdown
+++ b/docs/warning_types/template_injection/index.markdown
@@ -1,0 +1,5 @@
+User input passed into ruby templates that are evaluated is VERY dangerous, so this will always raise a warning. Brakeman looks foir calls of the form:
+
+```ruby
+  ERB.new(user_input).result
+```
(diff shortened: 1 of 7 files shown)


presidentbeef:
That's fine, but let's introduce a new `warning_code` and `warning_type` for this. I'd suggest `:template_injection` and `"Template Injection"` but I am open.

mhickman:
@presidentbeef done

presidentbeef:
Thanks Matt.

Let's work on the message a bit... I think it should be a little bit more specific and also include the type of template.

Something like

```ruby
msg(msg_input(input), " directly in ", msg_code("ERb"), " template, which might allow remote code execution")
```

`msg_input(input)` will output a friendly version of the value type (e.g. "parameter value").

Since this is going to include other template libraries in the future, I think it would be nice to parameterize on the template library name.

I think the message should also include some information about why it is dangerous. The old `eval` message is not a good example :) Above is a suggestion but I'd appreciate your input as well.

presidentbeef:
@mhickman any thoughts on the warning message?

PR #1400, opened 2019-09-23
Added direct line to file

dukaev:
With direct line it will be faster to open the file

For example https://www.dropbox.com/s/caxgsfhw5ftpxp4/sublime.mp4?dl=0

### DIFF
--- a/lib/brakeman/report/ignore/interactive.rb
+++ b/lib/brakeman/report/ignore/interactive.rb
@@ -282,7 +282,14 @@ q - Quit, do not update ignored warnings
 
       if warning.file
         label "File"
-        say warning.file.relative
+
+        file = if warning.line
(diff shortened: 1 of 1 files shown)


dukaev:
@presidentbeef hello!
What do you think about it?

presidentbeef:
Hi @dukaev - thanks for opening this PR.

The line number is already included in the output, though, so I think it would be weird to add it to the file name:
Screen Shot 2019-09-23 at 11 20 27 AM

dukaev:
I agree. But it would make it easier to navigate the code

dukaev:
@presidentbeef and what about now https://github.com/presidentbeef/brakeman/pull/1400/commits/21f05f72fdf75fea840dc9102d9fd47a8e9363b9 ? 😂

PR #1406, opened 2019-10-15
Refactor Brakeman::Differ#second_pass

Becojo:
## Issue
I had noticed that `Brakeman::Differ` can sometimes provide inaccurate results (for example:  missing new warnings or incorrectly reporting a warning as new) but I haven't really dug into the code until today.

Turns out the issue is in `Brakeman::Differ#second_pass` when it attempts to delete elements using indices while iterating the lists:
https://github.com/presidentbeef/brakeman/blob/5633686ddf653a4a811638e1ed6b54036d2e23b8/lib/brakeman/differ.rb#L38-L40

In my case, the output of both `delete_at` were not the same warning. This caused the fixed list to become empty when the new list still had warnings in it. This would cause the remaining warnings to be considered as new. 

## Solution

I've redone the algorithm by building a set of fingerprints from both warning lists which I then intersect to remove in bulk from both lists. It should have a better complexity than the previous algorithm.

Though, I left a similar algorithm afterwards for backward compatibility with old warning keys (I'm not sure if it's really needed).

Also, I renamed `DifferTests#diff` to `DifferTests#run_diff` because it shadowed `Minitest::Test#diff` which was quite confusing while coding this.

### DIFF
--- a/lib/brakeman/differ.rb
+++ b/lib/brakeman/differ.rb
@@ -24,43 +24,31 @@ class Brakeman::Differ
   # second pass to cleanup any vulns which have changed in line number only.
   # Given a list of new warnings, delete pairs of new/fixed vulns that differ
   # only by line number.
-  # Horrible O(n^2) performance.  Keep n small :-/
   def second_pass(warnings)
-    # keep track of the number of elements deleted because the index numbers
-    # won't update as the list is modified
-    elements_deleted_offset = 0
+    new_fingerprints = Set.new(warnings[:new].map(&method(:fingerprint)))
+    fixed_fingerprints = Set.new(warnings[:fixed].map(&method(:fingerprint)))
 
-    # dup this list since we will be deleting from it and the iterator gets confused.
-    # use _with_index for fast deletion as opposed to .reject!{|obj| obj == *_warning}
-    warnings[:new].dup.each_with_index do |new_warning, new_warning_id|
-      warnings[:fixed].each_with_index do |fixed_warning, fixed_warning_id|
-        if eql_except_line_number new_warning, fixed_warning
-          warnings[:new].delete_at(new_warning_id - elements_deleted_offset)
-          elements_deleted_offset += 1
-          warnings[:fixed].delete_at(fixed_warning_id)
-          break
-        end
+    # Remove warnings which fingerprints are both in :new and :fixed
+    shared_fingerprints = new_fingerprints.intersection(fixed_fingerprints)
+
+    unless shared_fingerprints.empty?
+      warnings[:new].delete_if do |warning|
+        shared_fingerprints.include?(fingerprint(warning))
+      end
+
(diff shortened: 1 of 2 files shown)


lib/brakeman/differ.rb
```diff
@@ -26,26 +26,54 @@ def diff
...
-    # won't update as the list is modified
-    elements_deleted_offset = 0
+    new_fingerprints = Set.new(warnings[:new].map(&method(:fingerprint)))
+    fixed_fingerprints = Set.new(warnings[:fixed].map(&method(:fingerprint)))
 
-    # dup this list since we will be deleting from it and the iterator gets confused.
-    # use _with_index for fast deletion as opposed to .reject!{|obj| obj == *_warning}
-    warnings[:new].dup.each_with_index do |new_warning, new_warning_id|
+    # Remove warnings which fingerprints are both in :new and :fixed
+    new_fingerprints.intersection(fixed_fingerprints).each do |hash|
```
presidentbeef: Would it be better to invert these loops?

I think the `delete_if` loops should be on the outside and they can check if the fingerprint is in the set.

Something like:

```ruby
shared_fingerprints = #...intersection
warnings[:new].delete_if do |warning
  shared_fingerprints.include? fingerprint(warning)
end
```
That way each array is only modified once.

You could also exit early if the intersection is empty.

lib/brakeman/differ.rb
```diff
@@ -26,26 +26,54 @@ def diff
...
+        fingerprint(warning) == hash
+      end
+
+      warnings[:fixed].delete_if do |warning|
+        fingerprint(warning) == hash
+      end
+    end
+
+    # Most duplicate warnings should be removed by now.
+    # The following code is to support old warning keys.
```
presidentbeef: I believe all the code dealing with "old keys" can be removed now. No one should have reports without "fingerprints".

lib/brakeman/differ.rb
Becojo: Good catch, thanks!  😅

Becojo:
🤦‍♂I misidentified the cause of my issue. So more context on my use case: I store the JSON reports and compare them using `Brakeman::Differ`. 

The issue only happens if the keys are not converted to back into symbols because the following check will return false:


https://github.com/presidentbeef/brakeman/blob/c9fc366db717f759c064371a0321cccc88c2affb/lib/brakeman/differ.rb#L56

Which then goes into the old warning key check

https://github.com/presidentbeef/brakeman/blob/c9fc366db717f759c064371a0321cccc88c2affb/lib/brakeman/differ.rb#L59-L61

Where the warnings are considered equal because every key is nil (because the keys are string and not symbols). 
---

In any case, we can probably repurpose this PR as a performance refactor of the Differ then. What do you think? I've address your comment about the loops and removed code about the old warning keys.

presidentbeef:
Thanks! :sun_with_face:

PR #1411, opened 2019-10-17
Catch reverse tabnabbing with :_blank symbol

JacobEvelyn:
Using `target: :_blank` in ERB and Haml also results in `target="_blank"` in HTML, so our reverse tabnabbing check should catch this case too.


presidentbeef:
Thanks!

PR #1412, opened 2019-10-17
Sort text report output by file and line

JacobEvelyn:
When Brakeman gives a lot of output, I find it can be really confusing to scroll through it and fix everything when its output is jumping between files seemingly at random. This is especially true where a new version of brakeman finds many new warnings, and the same warning occurs multiple times within the same file. Because Brakeman's output is not sorted by file/line, it appears as though Brakeman "missed" some warnings in the file until you scroll through its entire output.

This commit simply has Brakeman include filename and line number in its sorting so output containing lots of warnings appears more coherent.

### DIFF
--- a/lib/brakeman/report/report_text.rb
+++ b/lib/brakeman/report/report_text.rb
@@ -19,7 +19,7 @@ class Brakeman::Report::Text < Brakeman::Report::Base
     add_chunk generate_controllers if tracker.options[:debug] or tracker.options[:report_routes]
     add_chunk generate_templates if tracker.options[:debug]
     add_chunk generate_obsolete
-    add_chunk generate_errors 
+    add_chunk generate_errors
     add_chunk generate_warnings
   end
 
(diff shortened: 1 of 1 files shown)


lib/brakeman/report/report_text.rb
```diff
@@ -19,7 +19,7 @@ def generate_report
     add_chunk generate_controllers if tracker.options[:debug] or tracker.options[:report_routes]
     add_chunk generate_templates if tracker.options[:debug]
     add_chunk generate_obsolete
-    add_chunk generate_errors
```
JacobEvelyn: Whoops, I forgot that my editor trims trailing whitespace. Let me know if you want me to remove these edits from the commit!

presidentbeef: Not a problem, I generally clean up as I go, too.

presidentbeef:
Yes, this makes sense to me.

Also keep in mind you can compare JSON reports with `--compare`.

Issue #1409, opened 2019-10-16
Flaky failing test


presidentbeef:
This test has been "randomly" failing for a long time. Someone should look into it...

```
   1) Failure: Rails4Tests#test_external_checks [/home/circleci/project/test/tests/rails4.rb:1670]: Expected nil to be truthy.
```

PR #1419, opened 2019-10-28
Fix flaky rails4 test

adamkiczula:
This fixes issue https://github.com/presidentbeef/brakeman/issues/1409

I found the failing seed with: `while rake; do :; done`
Failure can be reproduced with `rake TESTOPTS="--seed=42281"`

The cause is that the `test_external_checks` test is run first, so `report` hasn't been called yet. The result is that `Brakeman::CheckExternalCheckTest` isn't loaded yet, so the `defined?` assertion fails. Simply moving that assertion to after calling `report` fixes the test.

### DIFF
--- a/test/tests/rails4.rb
+++ b/test/tests/rails4.rb
@@ -1667,8 +1667,8 @@ class Rails4Tests < Minitest::Test
 
   #Verify checks external to Brakeman are loaded
   def test_external_checks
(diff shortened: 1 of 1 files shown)


presidentbeef:
:bowing_man: :bowing_man: :bowing_man: Thank you!!

PR #1430, opened 2019-11-12
Adding git blame in that warnings

honestveera:
## Subject
Add Git blame for brakeman warnings!!!!
## Description
We have all the informations in  warnings.
but, we missed Who has done unsecured code? . so, I added `git blame` code.
## Summary of CodeChanges
` lib/brakeman/warning.rb`
## Reviewer
  - [ ] @presidentbeef 
  - [ ] @brynary

### DIFF
--- a/lib/brakeman/warning.rb
+++ b/lib/brakeman/warning.rb
@@ -289,6 +289,7 @@ class Brakeman::Warning
       :message => self.message.to_s,
       :file => (absolute_paths ? self.file.absolute : self.file.relative),
       :line => self.line,
+      :blame => blame(absolute_paths),
       :link => self.link,
       :code => (@code && self.format_code(false)),
       :render_path => render_path,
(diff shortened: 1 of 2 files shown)


abhilash1985:
Nice one @honestveera

honestveera:
Hi @presidentbeef 
Please review my PR and update
Thanks

presidentbeef:
Hi @honestveera. Thank you for your effort in this PR.

However, I do not think this is a good feature for Brakeman. I think it's better to left to external tools to connect Brakeman reports/warnings with code changes and code reviews. I also do not believe it is useful to "blame" developers for specific warnings outside of a code review process.

Some feedback on the code change itself:

- The code assumes `git` is installed
- The code assumes the code being scanned is in a git repo
- It does not appear to work if Brakeman is not run from the root of the application
- The "blame" info is very messy in my opinion and not machine-readable
- I would recommend putting this behind a command line option instead of having it be default behavior

If you are interested in having this feature, I recommend releasing it as a separate gem. I do not think the behavior should be in Brakeman itself.

PR #1436, opened 2019-11-27
initialize rails version with nil

jethroo:
while running the test with `rake` I got lots of following warnings: 
```
/Users/carstenwirth/Workspace/brakeman/lib/brakeman/tracker/config.rb:120: warning: instance variable @rails_version not initialized
```
these can easily be silenced by initializing the ivar as done in this PR


presidentbeef:
Thanks!

PR #1439, opened 2019-12-20
Adding a user-friendly message when ignore config file has invalid JSON

deevis:
We had a case where the brakeman.ignore file was manually edited resulting in invalid JSON.  

It was more of an effort than it seems like it should have been to determine what had happened.


presidentbeef:
Awesome, thanks!

PR #1440, opened 2019-12-20
Catch dangerous concatenation in check_execute

### LINKED ISSUE #1197: Command injection warning expected but missed
### Background

Brakeman version: 4.2.1
Rails version: 5.1.6
Ruby version: 2.3.4

### Issue

Similar to #1179, I'd expect the following code to produce a command injection warning in brakeman:

```ruby
def dangerous(foo)
  system("echo " + foo)
end
```

Would it be possible to get brakeman to transform that into `system("echo #{foo}")`, similar to your suggested solution for #1179?

JacobEvelyn:
Closes #1197

### DIFF
--- a/lib/brakeman/checks/base_check.rb
+++ b/lib/brakeman/checks/base_check.rb
@@ -280,15 +280,6 @@ class Brakeman::BaseCheck < Brakeman::SexpProcessor
     return location, line
   end
 
-  #Checks if an expression contains string interpolation.
-  #
-  #Returns Match with :interp type if found.
-  def include_interp? exp
-    @string_interp = false
-    process exp
-    @string_interp
-  end
-
   #Checks if _exp_ includes user input in the form of cookies, parameters,
   #request environment, or model attributes.
   #
@@ -503,5 +494,17 @@ class Brakeman::BaseCheck < Brakeman::SexpProcessor
     end
 
     @active_record_models
+  end
+
+  STRING_METHODS = Set[:<<, :+, :concat, :prepend]
+  private_constant :STRING_METHODS
+
+  def string_building? exp
+    return false unless call? exp and STRING_METHODS.include? exp.method
+
+    node_type? exp.target, :str, :dstr or
+    node_type? exp.first_arg, :str, :dstr or
+    string_building? exp.target or
+    string_building? exp.first_arg
   end
 end
--- a/lib/brakeman/checks/check_execute.rb
+++ b/lib/brakeman/checks/check_execute.rb
@@ -56,8 +56,20 @@ class Brakeman::CheckExecute < Brakeman::BaseCheck
 
     case call.method
     when :popen
-      unless array? first_arg
-        failure = include_user_input?(args) || dangerous_interp?(args)
+      # Normally, if we're in a `popen` call, we only are worried about shell
+      # injection when the argument is not an array, because array elements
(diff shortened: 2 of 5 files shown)


lib/brakeman/checks/base_check.rb
```diff
@@ -280,15 +280,6 @@ def get_location result
     return location, line
   end
 
-  #Checks if an expression contains string interpolation.
-  #
-  #Returns Match with :interp type if found.
-  def include_interp? exp
```
JacobEvelyn: This was only used in `check_execute.rb` so I moved it there.

lib/brakeman/checks/base_check.rb
```diff
@@ -504,4 +495,16 @@ def active_record_models
 
     @active_record_models
   end
+
+  STRING_METHODS = Set[:<<, :+, :concat, :prepend]
```
JacobEvelyn: These were moved from `check_sql.rb`

JacobEvelyn:
I'm not able to see the diff coverage in Code Climate, but please let me know if you have any guidance for how I might improve that to get the PR check to pass.

As always, I'm of course open to any feedback on this PR in general! This part of the code was pretty foreign to me and I make no claims to have done anything in the optimal way 😄

presidentbeef:
Hi @JacobEvelyn! Thank you for your effort here. I'm on a plane heading to vacation at the moment, but I will definitely take a look when I get a chance.

For code coverage, you can install simplecov and run the tests. Or maybe try the Code Climate browser plugin. Accessing pull request coverage is always a little weird for some reason.

lib/brakeman/checks/check_execute.rb
```diff
@@ -56,8 +56,20 @@ def process_result result
...
+      # injection when the argument is not an array, because array elements
+      # are always escaped by Ruby. However, an exception is when the array
+      # contains two values are something like "bash -c" because then the third
+      # element is effectively the command being run and might be a malicious
+      # executable if it comes (partially or fully) from user input.
+      if !array?(first_arg)
+        failure = include_user_input?(first_arg) ||
+                  dangerous_interp?(first_arg) ||
+                  dangerous_string_building?(first_arg)
+      elsif dash_c_shell_command?(first_arg[1], first_arg[2])
```
JacobEvelyn: I realized I missed this case when implementing the `-c` check in a previous PR.

test/apps/rails5.2/lib/shell.rb
```diff
@@ -94,4 +94,38 @@ def interpolated_in_percent_W
     # Should not warn
     system(*%W(foo bar #{value}))
   end
+
+  def completely_external
+    system(foo) # We assume this is safe and do not warn.
```
JacobEvelyn: @presidentbeef can you confirm that this is the behavior we want?

JacobEvelyn:
@presidentbeef I made some changes to the PR and force-pushed it. The new version passes the simplecov checks.

lib/brakeman/checks/base_check.rb
presidentbeef: Okay. Might move back at some point :)

test/apps/rails5.2/lib/shell.rb
presidentbeef: Yes. This check is already pretty noisy. If we don't know what the input is, let's not report.

PR #1457, opened 2020-02-13
Sort by fingerprint and line number for ignorefile

ngan:
When fingerprints are the same, the sorting is not deterministic and causes diffs in ignorefile when it gets updated. Add line number to the sorting algo as a tie breaker.

This is done here as well:
https://github.com/presidentbeef/brakeman/blob/master/lib/brakeman/report/report_json.rb#L41


presidentbeef:
Thank you so much for your work on adding this report format! 🙇

PR #1471, opened 2020-03-31
Add link to Brakeman github action

artplan1:
Add link to simple GH action that runs brakeman


presidentbeef:
There are a couple Actions available to run Brakeman, e.g. https://github.com/marketplace/actions/brakeman-linter. I'd rather not endorse a specific one. Maybe we can link to https://github.com/marketplace?type=actions&query=brakeman ?

presidentbeef:
Thanks!

PR #1478, opened 2020-05-12
Fix authenticate_or_request_with_http_basic check for passed blocks

hugocorbucci:
### Background

Brakeman version: 4.8.0
Rails version: 4.2.11
Ruby version: 2.5.8

### Issue

Running brakeman against a rails application with a method that calls `authenticate_or_request_with_http_basic` without a non direct call (as in the test).

Stack trace:

```
== Errors ==
Error: undefined method `last' for nil:NilClass
Location: /usr/local/bundle/gems/brakeman-4.8.0/lib/brakeman/checks/check_basic_auth.rb:60:in `include_password_literal?'
```


presidentbeef:
Thank you!!

PR #1489, opened 2020-06-25
Add --ensure-ignore-notes flag

eliblock:
Add a command line option, `--ensure-ignore-notes`, which causes Brakeman to exit with an error if the used ignore file contains any ignored warnings with blank notes.

This flag re-checks the ignore file after all other filtering operations, so it is safe to use in conjunction with `-I`.

This flag has no effect when used in conjunction with `--compare`.

### DIFF
--- a/lib/brakeman.rb
+++ b/lib/brakeman.rb
@@ -502,17 +502,16 @@ module Brakeman
     end
   end
 
-  # Returns false if the specified ignore file exists and contains an ignored
-  # warning without a note.
-  # Returns true otherwise.
-  def self.all_ignore_file_entries_have_notes? file
-    return true unless file
+  # Returns an array of alert fingerprints for any ignored warnings without
+  # notes found in the specified ignore file (if it exists).
+  def self.ignore_file_entries_with_empty_notes file
+    return [] unless file
 
     require 'brakeman/report/ignore/config'
 
     config = IgnoreConfig.new(file, nil)
     config.read_from_file
-    config.already_ignored_entries_with_empty_notes.empty?
+    config.already_ignored_entries_with_empty_notes.map { |i| i[:fingerprint] }
   end
 
   def self.filter_warnings tracker, options
--- a/lib/brakeman.rb
+++ b/lib/brakeman.rb
@@ -512,7 +512,7 @@ module Brakeman
 
     config = IgnoreConfig.new(file, nil)
     config.read_from_file
-    config.already_ignored_entries_without_notes.empty?
+    config.already_ignored_entries_with_empty_notes.empty?
   end
 
   def self.filter_warnings tracker, options
--- a/lib/brakeman.rb
+++ b/lib/brakeman.rb
@@ -508,6 +508,8 @@ module Brakeman
   def self.all_ignore_file_entries_have_notes? file
     return true unless file
 
+    require 'brakeman/report/ignore/config'
+
     config = IgnoreConfig.new(file, nil)
     config.read_from_file
     config.already_ignored_entries_without_notes.empty?
--- a/lib/brakeman.rb
+++ b/lib/brakeman.rb
@@ -20,6 +20,10 @@ module Brakeman
   #option is set
   Errors_Found_Exit_Code = 7
 
+  #Exit code returned when an ignored warning has no note and
+  #--ensure-ignore-notes is set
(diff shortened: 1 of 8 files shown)


lib/brakeman/options.rb
```diff
@@ -325,7 +329,7 @@ def create_option_parser options
           options[:min_confidence] =  3 - level.to_i
         end
 
-        opts.on "--compare FILE", "Compare the results of a previous Brakeman scan (only JSON is supported)" do |file|
+        opts.on "--compare FILE", "Compare the results of a previous Brakeman scan (only JSON is supported). Deactivates --ensure-ignore-notes" do |file|
```
presidentbeef, line 332: Instead of documenting here (which makes the help line pretty long), what about detecting that two conflicting options were given?

That could be added in `Commandline#parse_options`.

lib/brakeman/commandline.rb
```diff
@@ -124,6 +124,12 @@ def regular_report options
         if tracker.options[:exit_on_error] and tracker.errors.any?
           quit Brakeman::Errors_Found_Exit_Code
         end
+
+        if tracker.options[:ensure_ignore_notes] and
+           not Brakeman::all_ignore_file_entries_have_notes? tracker.ignored_filter&.file
```
presidentbeef, line 129: Add a space after the condition, please. Otherwise my brain hurts :laughing:

lib/brakeman.rb
```diff
@@ -498,6 +502,19 @@ def self.load_brakeman_dependency name, allow_fail = false
     end
   end
 
+  # Returns false if the specified ignore file exists and contains an ignored
+  # warning without a note.
+  # Returns true otherwise.
+  def self.all_ignore_file_entries_have_notes? file
```
presidentbeef, line 508: If this logic were moved to `Brakeman::Commandline` then the error message could list which warnings (fingerprints?) are missing notes.

This method could remain the same except for the call to `empty?`.

lib/brakeman/report/ignore/config.rb
```diff
@@ -94,6 +94,10 @@ def prune_obsolete
       end
     end
 
+    def already_ignored_entries_without_notes
+      @already_ignored.map { |i| i if !i[:note] || i[:note] == '' }.compact
```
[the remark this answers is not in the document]
presidentbeef: `select` may be a better choice here than `map {...}.compact`.

lib/brakeman.rb
eliblock: New sample error message:
```
Error: notes required for all ignored warnings when --ensure-ignore-notes is set. No notes provided for these alerts: ["734d17c982165a84df11282ba1f87ead6017df7be9d2b44c816dfae0a51fb8ac"]
```

lib/brakeman/options.rb
eliblock: I structured this so that we fail in favor of the user's likely intention, and simply de-activate `--ensure-ignore-notes` if it is used in conjunction with `--compare`, outputting the following warning (and then continuing to process the comparison):
```
Warning: --ensure-ignore-notes may not be used at the same time as --compare. Deactivating --ensure-ignore-notes. Please see `brakeman --help` for valid options
```

Let me know if you think it would be better to exit with an error in this case!

lib/brakeman/commandline.rb
```diff
@@ -124,6 +131,18 @@ def regular_report options
...
+
+        if tracker.options[:ensure_ignore_notes]
+          fingerprints = Brakeman::ignore_file_entries_with_empty_notes tracker.ignored_filter&.file
+
+          unless fingerprints.empty?
+            quit Brakeman::Empty_Ignore_Note_Exit_Code,
+                 'Error: notes required for all ignored warnings when ' \
+                 '--ensure-ignore-notes is set. No notes provided for these ' \
+                 'alerts: ' \
+                 "#{fingerprints}"
```
presidentbeef, line 143: Some tweaks here:

I think `[Error]` or `[Notice]` is more inline with Brakeman's messages. Also capitalized `Notes` looks better.

`alerts` should be `warnings` or `fingerprints`

The message is pretty long. Maybe add a newline before the fingerprints? I'd consider one fingerprint per line, too, like the obsolete fingerprints in reports.

PR #1491, opened 2020-06-30
Rev slim dependency to 4.1 to fix colon & slash issues

### LINKED ISSUE #1490: Slim parsing error - valid slim - slashes
### Background

Brakeman version: 4.8.2
Rails version: 6.0.3.2
Ruby version: 2.6.6
Slim 4.1.0

example slim code that is valid and works in-app but not when brakeman is run:
`.pr-2.w-1/2`

slash support was recently added to slim [here](https://github.com/slim-template/slim/pull/841)

change it to `.pr-2 class='w-1/2'` and brakeman will succeed

#### Parse Error

```
== Errors ==

Error: Unexpected text after closed tag   /home/jchurch/project/app/views/user_account_links/show.html.slim, Line 14, Column 13     .w-1/2.pr-2          ^ 
Location: /home/jchurch/project/app/views/user_account_links/show.html.slim:14
```

Minimal example that does not parse:

```slim
.w-1/2
  | example
```

If there is a better place to I should raise this to please point me in the right direction - I saw haml and erb on the parser error page but not slim.

Thank you so much!

JeremiahChurch:
Fixes #1490

### DIFF
--- a/gem_common.rb
+++ b/gem_common.rb
@@ -20,7 +20,7 @@ module Brakeman
       spec.add_dependency "highline", "~>2.0"
       spec.add_dependency "erubis", "~>2.6"
       spec.add_dependency "haml", "~>5.1"
-      spec.add_dependency "slim", ">=1.3.6", "<=4.0.1"
+      spec.add_dependency "slim", ">=1.3.6", "<=4.1"
     end
   end
 end


presidentbeef:
Thanks @JeremiahChurch !

PR #1493, opened 2020-07-10
Add check for CVE-2020-8166

chair6:
This adds a simple check for the CSRF token forgery issue, CVE-2020-8166 (https://groups.google.com/g/rubyonrails-security/c/NOjKiGeXUgw).

It's based purely on Rails version #s & I used https://github.com/presidentbeef/brakeman/blob/main/lib/brakeman/checks/check_basic_auth_timing_attack.rb for inspiration.

### DIFF
--- a/lib/brakeman/checks/check_csrf_token_forgery_cve.rb
+++ b/lib/brakeman/checks/check_csrf_token_forgery_cve.rb
@@ -1,0 +1,28 @@
+require 'brakeman/checks/base_check'
+
+class Brakeman::CheckCSRFTokenForgeryCVE < Brakeman::BaseCheck
+  Brakeman::Checks.add self
+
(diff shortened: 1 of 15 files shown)


presidentbeef:
> actually include the new check...

:laughing: I have done this so many times.

Thank you for contributing this check! FYI, I have not been trying to add checks for every CVE because dependency vulnerability checking is usually better handled by software composition analysis tools. But I don't mind including this one since you have done the work.

I will take a look.

chair6:
Ahh, that makes sense.. happy for you to decline-to-merge if you want to stick to that policy. Thanks for checking it out!

presidentbeef:
Can you also add a test to `test-cves.rb` for Rails 6? The current test app is on a beta version which doesn't trigger this CVE.

(Alternatively, if you want, you can update the test app to 6.0.0, but I'm not sure what all that would break.)

PR #1506, opened 2020-08-19
Also track active_record for version detection

### LINKED ISSUE #1494: Catching version when using `ActiveRecord` without rails!
Hi,

And thanks for the great work 🎸

We've been working with brakeman a lot in our team. However, we're mostly using sinatra with activerecord and there are some issues with the rails version we are using.

This lead us to having report for issues that are no longer present since rails 4.

Since activerecord and rails share their versions, i'd suggest appending `gem_version(:activerecord)` to the next line:

https://github.com/presidentbeef/brakeman/blob/d6421d4878ba7b13b5c52dca051360d8b4d1e3eb/lib/brakeman/tracker/config.rb#L82

This would be a great benefit for all activerecord users IMHO.

What do you think?

BuonOmo:
Since a lot of brakeman's report are related to AR (for instance, SQLI),
we track AR's version so brakeman can be easier to use in projects that
do not include rails. Sinatra applications for instance.

Closes #1494.

### DIFF
--- a/lib/brakeman/tracker/config.rb
+++ b/lib/brakeman/tracker/config.rb
@@ -79,7 +79,9 @@ module Brakeman
                   # Only used by Rails2ConfigProcessor right now
                   extract_version(version)
                 else
-                  gem_version(:rails) || gem_version(:railties)
+                  gem_version(:rails) ||
+                    gem_version(:railties) ||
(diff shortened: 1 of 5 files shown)


presidentbeef:
Thanks!

PR #1521, opened 2020-10-06
Fix less invocation for busybox implementation

jacekszubert:
Busybox implementation of less doesn't support -? parameter, which
results in redirecting help to stderr rather than stdout. Parameters in
help are separated from their description with tab rather than space.

### DIFF
--- a/lib/brakeman/report/pager.rb
+++ b/lib/brakeman/report/pager.rb
@@ -90,10 +90,10 @@ module Brakeman
       @less_options = []
 
       if system("which less > /dev/null")
-        less_help = `less -?`
+        less_help = `less -? 2>&1`
 
-        ["-R ", "-F ", "-X "].each do |opt|
-          if less_help.include? opt
+        ["-R", "-F", "-X"].each do |opt|
+          if less_help.match(opt + '\s')
             @less_options << opt
           end
         end


lib/brakeman/report/pager.rb
```diff
@@ -90,10 +90,10 @@ def less_options
       @less_options = []
 
       if system("which less > /dev/null")
-        less_help = `less -?`
+        less_help = `less -? 2>&1`
 
-        ["-R ", "-F ", "-X "].each do |opt|
-          if less_help.include? opt
+        ["-R", "-F", "-X"].each do |opt|
+          if less_help.match(opt + '\s')
```
presidentbeef, line 96: Needs to be updated in the `set_color` method, too, to match without a space.

presidentbeef (changes requested):
Hi @jacekszubert thanks for looking into this!

Weirdly, though, for me it works fine without your changes? What was the behavior you were seeing?

I started busybox like this:

```
 $ busybox sh

BusyBox v1.27.2 (Ubuntu 1:1.27.2-2ubuntu3.3) built-in shell (ash)
```

thomasLeclaire:
Hello! @jacekszubert you closed without merge it, does it mean you found a solution? 
I'm currently experimenting such issue also on my side. 

image

Issue #1387, opened 2019-07-23
Namespaced classes that are not fully qualified can cause difference in false positives/negatives


doliveirakn:
### Background

Brakeman version: 4.5.1
Rails version:  5.0.7.2
Ruby version: 2.4.3

Link to Rails application code:

### Issue

What problem are you seeing?

If we reference a namespaced ActiveRecord module without it being fully qualified constant, Brakeman will fail to identify it as a model and may report false positives or negatives. 

Code:
```ruby
# app/models/document.rb
class Document < ApplicationRecord
  attr_accessor :owner
end

# app/models/namespace/task.rb
module Namespace
  class Task < ApplicationRecord
    attr_accessor :owner
  end
end

# app/controllers/documents_controller.rb
class DocumentsController < ApplicationController
  def index
    redirect_to Document.new(params.permit(:owner)).owner
  end

  def show
    redirect_to Document.find(params[:id])
  end
end

# app/controllers/namespace/tasks_controller.rb
module Namespace
  class TasksController < ApplicationController
    def index
      redirect_to Task.new(params.permit(:owner)).owner
    end

    def show
      redirect_to Task.find(params[:id])
    end
  end
end
```

Running brakeman on this will result in 2 security warnings
```
== Warnings ==

Confidence: High
Category: Redirect
Check: Redirect
Message: Possible unprotected redirect
Code: redirect_to(Document.new(params.permit(:owner)).owner)
File: app/controllers/documents_controller.rb
Line: 4

Confidence: Weak
Category: Redirect
Check: Redirect
Message: Possible unprotected redirect
Code: redirect_to(Task.find(params[:id]))
File: app/controllers/namespace/tasks_controller.rb
Line: 9
```

It appears that brakeman is able to determine that `Document` is an ActiveRecord model but it cannot determine that `Task` is a model. This causes the checks to differ in behaviour and may report false negatives (`Namespace::TasksController#index` should have the same warning as the document one), or false positives (`Namespace::TasksController#show` should not have this warning)

presidentbeef:
Thank you for the clear examples and the example application!

I am looking at this now... I really don't want to try to replicate Ruby's name resolution, but I don't think it would be too difficult to at least store alternate names for each class.

gavingmiller:
I've tracked this down with the code that @doliveirakn included for his example app to the problem occurring in:

[...]

PR #1523, opened 2020-10-16
Namespaced classes that are not fully qualified can cause difference in false positives/negatives (WIP)

ChrisNJ58:
Hello!

A coworker and I have recently been working on trying to resolve an issue reported into Brakeman a little over a year ago, related to the namespacing of classes causing potential false negatives. There was an issue brought up related to this issue [here](https://github.com/presidentbeef/brakeman/issues/1387) if a refresher on the topic is needed.

This PR contains a fix that you had been [working on yourself](https://github.com/presidentbeef/brakeman/compare/class_names), before eventually putting aside, due to the results coming back with a number of false positives and weird results, specifically when analyzing controllers. I've taken your fix and put it on top the latest `4.10.0` version of Brakeman, including fixing a few tests that had been failing. 

When I and my coworker have been testing this fix, we do still some false positive results when running against some of our repositories (particularly where a number of previously ignored warnings are being re-raised, due to this fix changing the confidence levels on warnings), but overall it looks like the changes made in this PR resolves the greater namespacing issue. I know it's likely been a long time since this problem was looked at, but I was hoping to get some of your insight into what other issues you were experiencing when you initially worked on this PR, particularly around weird results you were finding on controller, and see if I could pick your brain on possible ideas that I could implement to get this problem solved!

### DIFF
--- a/lib/brakeman/checks/base_check.rb
+++ b/lib/brakeman/checks/base_check.rb
@@ -432,14 +432,12 @@ class Brakeman::BaseCheck < Brakeman::SexpProcessor
   #the current tracker, or else @models should contain an array of the model
   #names, which is available via tracker.models.keys
   def model_name? exp
-    @models ||= @tracker.models.keys
-
     if exp.is_a? Symbol
-      @models.include? exp
+      tracker.models.include? exp
     elsif call? exp and exp.target.nil? and exp.method == :current_user
       true
     elsif sexp? exp
-      @models.include? class_name(exp)
+      tracker.models.include? class_name(exp)
     else
       false
     end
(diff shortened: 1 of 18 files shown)


presidentbeef:
Whew, thanks for picking this back up! I will have to run some tests to remember what the issues were that I saw before.

ChrisNJ58:
@presidentbeef That sounds great! We've been running Brakeman with this fix on top on a number of our projects, and haven't been able to find anything strange or concerning from our results, so I'm really interested to hear if there's something that we're missing, and getting a chance to dig into it a bit deeper!

kfinn-clio:
One thing my co-worker and I noticed with this namespace change is previous ignored warnings become obsolete. After investigation these obsolete warnings are still valid but can be ignored. The obsolete warnings were in both models and controllers. Had these warnings not been ignored they would have not been flagged with the namespace change. 
Is this something you noticed when previously working on this change?

ChrisNJ58:
Hello again! 

I was wondering if you have had the chance to get reacquainted with this namespacing issue? It looks like since I posted this PR it's no longer passing codeclimate, but I would like to hear your thoughts if it would resolve the potential namespacing problem.

presidentbeef:
Hi folks, my priority right now is getting Brakeman 5.0 out. I apologize for the lack of reply.

PR #1531, opened 2020-11-27
Handle config settings like config.load_defaults(...)

presidentbeef:
Not just attribute settings like `config.blah = ...`


presidentbeef:
Closing in favor of #1532 which includes this commit.

PR #1575, opened 2021-03-02
Expand the configuration docs to include an example

### LINKED ISSUE #1574: Missing docs on how to add brakeman config
I need to ignore a directory in my project. I'd like this directory to always be ignore in the default configuration file so that I can run `brakeman` locally, without having to specify `brakeman --skip-files ...`

I have no idea how to write a brakeman config file. I've found in the readme that I can create `config/brakeman.yml`, but no idea on how the contents of that file should look like. It would be much easier to read the documentation than to write some valid YAML hoping I guessed correctly how `--skip-files` translates into YAML.

Did I miss the docs for this somewhere?

Thank you!

EDIT:
After digging through the source code, I managed to get it to work:

```yaml
skip_files:
  - 'plugins/'
```

The keys are underscored and the value (in this case at least) has to be an array. Just passing a single string results in an error.

vfonic:
Closes #1574

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,16 @@ The `-w` switch takes a number from 1 to 3, with 1 being low (all warnings) and 
 
 # Configuration files
 
-Brakeman options can stored and read from YAML files. To simplify the process of writing a configuration file, the `-C` option will output the currently set options.
+Brakeman options can be stored and read from YAML files.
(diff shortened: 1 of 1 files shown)


presidentbeef:
Thanks!

Issue #1529, opened 2020-11-13
Renderables dynamic render path false positive


bradparker:
Firstly, thanks so much for this tool. We're ([Five Good Friends](https://github.com/fivegoodfriends/)) grateful to have something like this to help us make sure we're doing all we can to keep the people who use our software safe :).

### Background

Brakeman version: 4.10
Rails version: 6.0.3.4
Ruby version: 2.6.6

Link to Rails application code: Sorry, closed-source. But I've added [a failing spec in a Brakeman fork](https://github.com/bradparker/brakeman/commit/d748f7bf03f0188e4dbc8daad469615fadd3b837)

#### False Positive

*Full* warning from Brakeman:

```json
{
  "warning_type": "Dynamic Render Path",
  "warning_code": 15,
  "fingerprint": "9485294198b3a144dd293dc9f26cc7653da664e5e6d79d15fcac79750a4b0228",
  "check_name": "Render",
  "message": "Render path contains parameter value",
  "file": "/Users/brad.parker/Code/presidentbeef/brakeman/test/apps/rails6/app/controllers/groups_controller.rb",
  "line": 12,
  "link": "https://brakemanscanner.org/docs/warning_types/dynamic_render_path/",
  "code": "render(action => TestRenderer.new(params.require(\"name\")), {})",
  "render_path": null,
  "location": {
    "type": "method",
    "class": "GroupsController",
    "method": "render_commands"
  },
  "user_input": "params.require(\"name\")",
  "confidence": "Weak"
}
```

Relevant code:

```ruby
def render_commands
  render(TestRenderer.new(params.require('name')))
end
```

_Why might this be a false positive?_

[Rails 6.1 now supports passing `Renderable`s to `render`](https://github.com/rails/rails/blob/b5b078cd99b0107fd8e31ff619578e5321182e0a/actionview/lib/action_view/template/renderable.rb). We've started using [Github's ViewComponent gem](https://github.com/github/view_component) which makes use of this new Rails feature.

I'm not sure what changes could be made to `CheckRender` but I'd be happy to take a swing at it with some pointers. It seems like there might be a way to add something similar to the `input.type == :model` check, is this on the right track?

Thanks again :)

presidentbeef:
Hi @bradparker! Thank you for bringing this up.

[...]

PR #1578, opened 2021-03-18
Skip dynamic render path check for renderables

bradparker:
[Rails 6.1 now supports passing `Renderable`s to `render`](https://github.com/rails/rails/blob/b5b078cd99b0107fd8e31ff619578e5321182e0a/actionview/lib/action_view/template/renderable.rb).

A renderable is anything that responds to `render_in`. This means we can
now expect to see `render(SomeRenderableClass.new(some_parameter))`. A
new library for building Rails UIs out of components, ViewComponent, is
growing in popularity and makes use of this new `render_in` API.

Closes https://github.com/presidentbeef/brakeman/issues/1529.


lib/brakeman/checks/check_render.rb
```diff
@@ -94,4 +95,21 @@ def safe_param? exp
...
-end 
+
+  def renderable? exp
+    call?(exp) and
+      sexp?(exp.target) and
+      constant?(exp.target) and
+      method_defined?(class_name(exp.target), :render_in)
+  end
+
+  def method_defined? class_name, name, seen = []
```
bradparker, line 106: I'm not sure about where best to put this, it feels too general too be scoped to only checking render calls. I also wanted to search `#includes` for `render_in` but the collection was always empty for my test components. I could see includes being populated for models but not for libs ... I could look better into how to populate includes? However, without that I _think_ this should address the ViewComponent use case as it looks like `render_in` is defined on `ViewComponent::Base`.

presidentbeef:
Hi @bradparker! Wow thanks for taking a swing at this! 🙏 

I think I have some pending changes that will help with method lookup.

I can't look too deeply at this PR right now but will follow up in the next few days!

bradparker:
> I think I have some pending changes that will help with method lookup.

Rad, thanks.

> I can't look too deeply at this PR right now but will follow up in the next few days!

No problem :). I feel like I'm probably missing some things so I really appreciate you taking the time, when ever you might have it, to have a look.

bradparker:
Ah, right. So, silly me, I've checked this in our code base just now. I didn't realize that Brakeman doesn't index dependencies so this doesn't quite work. It looks up `ViewComponent::Base` but can't find it so doesn't identify ViewComponents as renderables ... perhaps it would be possible to configure some class names as "renderables" if you're not comfortable adding specific handling for ViewComponent?

bradparker:
https://github.com/presidentbeef/brakeman/pull/1578/commits/8fe395edeb000b6f170f7f7093d31c392f41b7cf is a perhaps a proof of concept for handling ViewComponent specially. I've plugged that into our code base and it's removed all the ViewComponent related 'dynamic render path' false positives.

lib/brakeman/checks/check_render.rb
```diff
@@ -94,4 +95,37 @@ def safe_param? exp
...
+      (known_renderable_class?(class_name(exp.target)) or method_defined?(class_name(exp.target), :render_in))
+  end
+
+  def known_renderable_class? start
+    traverse_ancestry(start) do |class_name|
+      return true if class_name == :"ViewComponent::Base"
+    end
+  end
+
+  def method_defined? start, name
```
presidentbeef, line 112: Check out `Tracker#find_method` I just merged from https://github.com/presidentbeef/brakeman/pull/1596 I think it takes care of this.

lib/brakeman/checks/check_render.rb
```diff
@@ -94,4 +95,37 @@ def safe_param? exp
...
+  def renderable? exp
+    call?(exp) and
+      sexp?(exp.target) and
+      constant?(exp.target) and
+      (known_renderable_class?(class_name(exp.target)) or method_defined?(class_name(exp.target), :render_in))
+  end
+
+  def known_renderable_class? start
+    traverse_ancestry(start) do |class_name|
+      return true if class_name == :"ViewComponent::Base"
```
presidentbeef, line 108: I think you can use `Collection#ancestor?` for this:

```ruby
klass = tracker.find_class(class_name)
klass.ancestor? :"ViewComponent::Base"
```

presidentbeef:
> [8fe395e](https://github.com/presidentbeef/brakeman/commit/8fe395edeb000b6f170f7f7093d31c392f41b7cf) is a perhaps a proof of concept for handling ViewComponent specially. I've plugged that into our code base and it's removed all the ViewComponent related 'dynamic render path' false positives.

That seems totally reasonable to me.

lib/brakeman/checks/check_render.rb
```diff
@@ -94,4 +95,17 @@ def safe_param? exp
       end
     end
   end
-end 
+
+  def renderable? exp
+    return false unless call?(exp) and sexp?(exp.target) and constant?(exp.target)
```
presidentbeef, line 100: `constant?` will check if it's a `Sexp` so you can skip the `sexp?` call.

lib/brakeman/checks/check_render.rb
```diff
@@ -94,4 +95,17 @@ def safe_param? exp
...
+
+    target_class_name = class_name(exp.target)
+    known_renderable_class?(target_class_name) or tracker.find_method(:render_in, target_class_name)
+  end
+
+  def known_renderable_class? class_name
```
presidentbeef, line 106: Can you add a test for this? Thanks!

presidentbeef (changes requested):

presidentbeef:
Thanks @bradparker!

bradparker:
No worries! Thanks :)

PR #1598, opened 2021-05-31
Adjust the copy of the interactive menu

elia:
The first and second options are referring to *inspecting warnings* and to *hiding warnings* respectively.
This is a bit confusing when the actual thing the user wants to do is to have a look at new warnings, and eventually decide if they need to be ignored.


presidentbeef:
Thanks @elia - that's much better!

PR #1604, opened 2021-06-07
Fix cve_2018_8048 false positive

tarhashi:
This PR fixes https://github.com/presidentbeef/brakeman/issues/1603#issue-912936927

String comparison of version strings sometimes makes false positive.
```
'2.10.0' < '2.2.1'
=> true
Gem::Version.new('2.10.0') < Gem::Version.new('2.2.1')
=> false
```

Issue #1394, opened 2019-08-20
Use rubygems to compare gem versions/requirements


presidentbeef:
Since rubygems is included in Ruby now, let's use [`Gem::Dependency`](https://www.rubydoc.info/github/rubygems/rubygems/Gem/Dependency) and/or [`Gem::Version`](https://www.rubydoc.info/github/rubygems/rubygems/Gem/Version) to compare gem versions instead of doing it manually.

PR #1605, opened 2021-06-07
Check loofah versions using Gem::Version instead of String#>

elia:
With loofah v2.10 ([just released](https://rubygems.org/gems/loofah/versions/2.10.0)) fails interpreting it as an earlier version than 2.2.

This should be ported to all version checks, but I thought this hotfix was still granted to avoid headaches and red CIs to other people facing the same issue.

### DIFF
--- a/lib/brakeman/checks/check_sanitize_methods.rb
+++ b/lib/brakeman/checks/check_sanitize_methods.rb
@@ -1,4 +1,5 @@
 require 'brakeman/checks/base_check'
+require 'rubygems/version'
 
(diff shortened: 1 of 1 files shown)


gaganawhad:
Thanks @elia.

presidentbeef:
This is identical to #1604

> This should be ported to all version checks

Most version checks use `version_between?`. Not really sure why this one didn't... laziness I guess.

But feel free to take a look at https://github.com/presidentbeef/brakeman/issues/1394

PR #1606, opened 2021-06-07
fix loofah version check

kbrock:
loofah is up to 2.10 which alphabetically is before 2.2
This changes the check to be a numerical comparison

Now it no longer has a false positive:

```
loofah gem 2.10.0 is vulnerable (CVE-2018-8048). Upgrade to 2.2.1
```

If you want me to go over the rest of the version checks, let me know.


presidentbeef:
Thanks @kbrock but this is identical to #1605 and #1604

PR #1608, opened 2021-06-07
merge cookie and param check

kbrock:
Hello,

Both cookie and param methods are recursive tests.
This extracts the recursion mechanism into a separate method.

Goal is to remove code climate warning:

https://codeclimate.com/github/presidentbeef/brakeman/lib/brakeman/util.rb/source#issue-25dd99aa71b8bc2260fef8f7a714191a

> Similar blocks of code found in 2 locations. Consider refactoring

As for testing, it looks like most of the `if` blocks are tested except for the last one, `if exp[2] == :[]`
Please double check that I didn't mess that one up.

### DIFF
--- a/lib/brakeman/util.rb
+++ b/lib/brakeman/util.rb
@@ -238,30 +238,22 @@ module Brakeman::Util
 
   #Check if _exp_ is a params hash
   def params? exp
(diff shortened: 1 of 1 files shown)


presidentbeef:
Thanks!

PR #1611, opened 2021-06-08
Fix infinite loop on mixin self-includes

Skipants:
Adds ae37c48:

> Fixes a regression in 5.0.2 where a module that had code including itself in a class method would cause Brakeman::Tracker#find_method to infinite loop.

Hi! Thanks for the gem! While upgrading from 5.0.1 to 5.0.2 we had this regression when running brakeman in our CI. I narrowed it down to this unorthodox-but-technically-valid-ruby. We have a module in our codebase that includes itself when building a generic class. It caused brakeman to infinite loop on `Brakeman::Tracker#find_method`. I turned the smallest PoC I could into a test case.

Another way to solve this might be to prevent a class being added to its own `includes` collection, but this was the easier fix. Thanks, again!

### DIFF
--- a/lib/brakeman/tracker/collection.rb
+++ b/lib/brakeman/tracker/collection.rb
@@ -24,7 +24,7 @@ module Brakeman
     def ancestor? parent, seen={}
       seen[self.name] = true
 
-      if self.parent == parent or seen[self.parent]
+      if self.parent == parent or self.name == parent or seen[self.parent]
         true
       elsif parent_model = collection[self.parent]
         parent_model.ancestor? parent, seen
@@ -39,7 +39,7 @@ module Brakeman
     end
 
     def add_include class_name
-      @includes << class_name
+      @includes << class_name unless ancestor?(class_name)
     end
 
     def add_option name, exp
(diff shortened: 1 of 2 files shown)


presidentbeef:
There shouldn't be any regression between 5.0.1->5.0.2. :thinking:  Can you try 5.0.4?

presidentbeef:
Hi @Skipants - thanks for the PR.

I think this should be more general. There are a couple other examples of avoiding infinite recursion loops in Brakeman, for example https://github.com/presidentbeef/brakeman/blob/9f2e4b4830bb2698fc11f024591a40d7b6686a99/lib/brakeman/tracker/collection.rb#L24-L34

Can you update this PR to use that approach? Thanks!

Skipants:
@presidentbeef I updated and rebased based on your suggestion. I think this is what you meant but let me know. I had to add the condition that a class is its own ancestor because that was the case I ran into. I hope that's not too heavy-handed of a change. All the tests pass for me locally.

presidentbeef:
Thanks!

PR #1612, opened 2021-06-10
Avoid false positive about Command Injection on Open3.capture

xulaus:
The last argument to `Open3.capture*` methods are execution options, one of which is a string to be used as the STDIN for the called process. It is useful to be able to use string interpolation on this argument and it is, in general, not succeptable to command injection.

Most of the other options are passed to [`Process.spawn`](https://ruby-doc.org/core-2.7.0/Process.html#method-c-spawn), and I don't believe any of those are succeptable either.

As such it is safe to ignore the last argument of the capture methods if it is a Hash.

### DIFF
--- a/lib/brakeman/checks/check_execute.rb
+++ b/lib/brakeman/checks/check_execute.rb
@@ -87,6 +87,16 @@ class Brakeman::CheckExecute < Brakeman::BaseCheck
                   dangerous_interp?(first_arg) ||
                   dangerous_string_building?(first_arg)
       end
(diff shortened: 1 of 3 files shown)


presidentbeef:
Thanks!

Not sure I 100% agree that passing in values to STDIN is "safe", but I guess it wouldn't be command injection.

> As such it is safe to ignore the last argument of the capture methods if it is a Hash.

Probably true... although `:chdir` seems a little dangerous.

Anyway - I'm fine with this change. Seems fairly rare to use it this way.

PR #1618, opened 2021-07-08
Initial support for ActiveRecord enums (and other things, too)

presidentbeef:
Tired of splitting this into separate PRs, so here we go!

For _very_ simple, obvious _class_ methods, Brakeman will now pull out the return value:

```ruby
class A
  def self.cool_beans
    "nice beans!"
  end
end

A.cool_beans # => "nice beans!"
```

This PR also builds on that to support use of [ActiveRecord `enum`](https://api.rubyonrails.org/classes/ActiveRecord/Enum.html), which essentially just defining a number of methods that return constant values or booleans.

```ruby
class B < ActiveRecord::Base
  enum status: [:waiting, :running, :done]
end

B.statuses # => { waiting: 0, running: 1, done: 2 }
```

Additionally:
* `pluralize` is slightly better now (`status` -> `statuses` instead of `statuss`)
* Support for `Hash#include?`

### DIFF
--- a/lib/brakeman/processors/model_processor.rb
+++ b/lib/brakeman/processors/model_processor.rb
@@ -73,6 +73,8 @@ class Brakeman::ModelProcessor < Brakeman::BaseProcessor
           @current_class.set_attr_accessible exp
         when :attr_protected
           @current_class.set_attr_protected exp
+        when :enum
+          add_enum_method exp
         else
           if @current_class
             @current_class.add_option method, exp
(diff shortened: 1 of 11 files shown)


lib/brakeman/processors/model_processor.rb
```diff
@@ -87,4 +89,33 @@ def process_call exp
...
+      enum_values = s(:hash).line(call_line)
+
+      enums.each_sexp.with_index do |v, index|
+        enum_values << v
+        enum_values << s(:lit, index).line(call_line)
+      end
+    end
+
+    enum_method = s(:defn, enum_name, s(:args), safe_literal(call_line))
+    enums_method = s(:defs, s(:self), enums_name, s(:args), enum_values)
```
alpaca-tc, line 116: I may be wrong but I think that `enum_values` have to take into consideration `_prefix/_suffix` options.
https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Enum.html

```ruby
enum(item_type: %i[digital]) #=> define digital
enum(item_type: %i[digital], _prefix: true) #=> define item_type_digital
```

alpaca-tc (commented):
Thank you for great PR 🎉 

I was a little worried 😄 🙏

PR #1620, opened 2021-07-13
feat: sarif output includes suppressed findings

eliblock:
Expand the SARIF report to include ignored warnings, properly annotated
with a [suppression](https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127924).

### DIFF
--- a/test/tests/sarif_output.rb
+++ b/test/tests/sarif_output.rb
@@ -102,7 +102,7 @@ class SARIFOutputTests < Minitest::Test
     assert_equal(
       1,
       @@sarif_with_ignore.dig('runs', 0, 'results').
-        map { |f| 1 if f['suppressions'] }.compact.sum
+        select { |f| f['suppressions'] }.count
     )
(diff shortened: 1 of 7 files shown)


test/tests/sarif_output.rb
```diff
@@ -96,4 +97,33 @@ def test_results_shape
       end
     end
   end
+
+  def test_with_ignore_has_one_suppressed_finding
+    assert_equal(
+      1,
+      @@sarif_with_ignore.dig('runs', 0, 'results').
+        map { |f| 1 if f['suppressions'] }.compact.sum
```
presidentbeef, line 105: `map { |f| 1 if f['suppressions'] }.compact.sum` :thinking: 

`select { f['suppressions'] }.count` ?

presidentbeef (changes requested):
One extremely minor comment

PR #1623, opened 2021-07-19
Revert "refactor: IgnoreConfig uses FilePath"

bradparker:
This reverts commit 52bfc9b187dcc898f631124bc841823de6117ca5.

It seems that this might break --interactive-ignore. Specifically on
line 39 of lib/brakeman/report/interactive.rb.

https://github.com/presidentbeef/brakeman/blob/53db6f585679f648ce761421a9eba2e02e18a4a1/lib/brakeman/report/ignore/interactive.rb#L39

When asking for the ignore file location `@ignore_config.file` is
checked for emptiness but Brakeman::FilePath has no `empty?` method and
so this fails.

### DIFF
--- a/lib/brakeman.rb
+++ b/lib/brakeman.rb
@@ -527,14 +527,12 @@ module Brakeman
 
   # Returns an array of alert fingerprints for any ignored warnings without
   # notes found in the specified ignore file (if it exists).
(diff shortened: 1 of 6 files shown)


bradparker:
@presidentbeef just in case this is useful. Understand if you'd prefer to roll-forward.

PR #1624, opened 2021-07-20
fix(InteractiveIgnorer): use FilePath

### LINKED ISSUE #1622: Interactive ignore tries to call `empty?` on a Brakeman::FilePath and fails
### Background

Brakeman version: 5.1.0
Rails version: 6.1.4
Ruby version: 2.7.3

Link to Rails application code: Sorry, it's a private repo but I'm not sure this is source related.

### Issue

When trying to run with `--interactive-ignore` I get the following error:

```
$BUNDLE_PATH/ruby/2.7.0/gems/brakeman-5.1.0/lib/brakeman/report/ignore/interactive.rb:39:in `block (2 levels) in file_menu': undefined method `empty?' for #<Brakeman::FilePath:0x00007fae2cbf4c08> (NoMethodError)
```

I _think_ this change may have introduced it: https://github.com/presidentbeef/brakeman/commit/52bfc9b187dcc898f631124bc841823de6117ca5.

eliblock:
Fixes #1622.

The refactor of ignore files to use FilePath in #1620 did not include
the InteractiveIgnorer. Complete the refactor.

The original PR included the refactor to FilePath in order to update
ignorefile pathing to match other pathing within Brakeman, which in turn
permitted the SARIF report to use the pre-loaded relative path in its
output.

As noted in #1622, this had the unintended side effect of making the
path of `-i` configurations relative to the app root (if an absolute
path was not provided), rather than relative to where the command was
run. This _differs_ from the behavior of the `-c` flag, may be
undesired, and is not addressed in this PR.

### DIFF
--- a/lib/brakeman.rb
+++ b/lib/brakeman.rb
@@ -556,7 +556,7 @@ module Brakeman
 
     if options[:interactive_ignore]
       require 'brakeman/report/ignore/interactive'
-      config = InteractiveIgnorer.new(file, tracker.warnings).start
+      config = InteractiveIgnorer.new(app_tree, file, tracker.warnings).start
     else
       notify "[Notice] Using '#{file}' to filter warnings"
       config = IgnoreConfig.new(file, tracker.warnings)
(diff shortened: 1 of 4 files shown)


eliblock:
#1623 reverted the cause of the issue.

PR #1629, opened 2021-07-27
Fix issue where the previous output is still visible 

Fryguy:
The output of the "libs" section was not fully clearing the previous line's output:

```text
Processing initializers...
Processing libs...ssed
Processing routes...
```

This first commit in this PR fixes that.  In addition, the second commit updates all of the other lines to be at full width for consistency, and also to prevent any future additions from accidentally occurring should someone copy/paste a new section.


Fryguy:
I don't believe my changes are the ones causing issues, but main is green, so I'm not sure what's changed here.

presidentbeef:
Tests are failing due to an unrelated change in HAML 👍

PR #1633, opened 2021-08-23
Add check for Host Authorization middleware

p8:
The Host Authorization middleware protects against DNS rebinding.
This middleware is primarily targeted at the development environment:

> It is included in the development environment by default ... In other
environments Rails.application.config.hosts is empty and no Host header
checks will be done.
https://github.com/rails/rails/pull/33145

If someone decides to call `config.hosts.clear` because it's "only
development", we should warn them they are vulnerable to DNS rebinding.


presidentbeef:
Hi @p8, thank you for putting this together!

I am pretty sure Brakeman doesn't even look at `config/development.rb`, though. Only `config/production.rb`. :thinking:

PR #1634, opened 2021-08-24
Update Docs for ruby version references

existere:
Minimum version has been updated to 2.4.0 here: https://github.com/presidentbeef/brakeman/pull/1515.


presidentbeef:
Fair enough. Thanks!

PR #1642, opened 2021-10-10
Various test fixes for new ruby_parser changes.

zenspider:
Mostly fixed by adding String#lineno=, which you should move away
from.

The rest were line numbers or error strings changing (also a tad
brittle to test since racc can change at any time).


presidentbeef:
Brakeman doesn't use `String#lineno=`, so that should probably be added in `ruby_parser-legacy`.

PR #1644, opened 2021-10-20
Update README.md - docker use with -o export

jympetalmd:


jympetalmd:
> > It would work if your current directory is writable by others (ex. 777)
> 
> I mean
> 
> ```
>  -v /tmp:/tmp \
> ```
> 
> and
> 
> ```
> -o /tmp/brakeman_results.html
> ```
> 
> parts, it must use /tmp as mount

Sorry I meant that without ANY extra volume mount, the original .md would work if the directory with code is writable by others. Didn't test it but pretty sure.

PR #1648, opened 2021-11-05
Fix error for endless method

mitsuru:


presidentbeef:
Thanks @mitsuru!

I think this is a ruby_parser bug...

```ruby
3.0.0 :003 > RubyParser.new.parse 'def x = y'
 => s(:defn, :x, nil, s(:call, nil, :y)) 
3.0.0 :004 > RubyParser.new.parse 'def x; y; end'
 => s(:defn, :x, s(:args), s(:call, nil, :y))
```

Those should not be different.

zenspider:
Agreed... Release with fix coming soon

zenspider:
released

presidentbeef:
Thanks @zenspider !

@mitsuru you can use this fix today by using the `brakeman-lib` gem and combining with the latest `ruby_parser`.

PR #1670, opened 2022-01-21
Update copy on unsafe reflection error message

pedropb:
`Unsafe reflection method constantize called with parameter value`
makes the reader think constantize was called with a parameter,
when in fact it was call on a parameter value.

Given these reflection methods usually don't take parameters, the message can be confusing.


presidentbeef (changes requested):
Can you update some tests to check the new message? (Not all of them... just one or two.)

For example: https://github.com/presidentbeef/brakeman/blob/6445afc8f925bb8f97daa32c6c657ef6c3f56bc5/test/tests/rails31.rb#L1178-L1195

pedropb:
> Can you update some tests to check the new message? (Not all of them... just one or two.)

You bet! Sorry for the delay.

I force pushed a new commit updating 2 assertions on each suite (rails2 and 31). Please let me know if this is what you had in mind.

presidentbeef:
Perfect, thanks!

Issue #1635, opened 2021-08-24
Brakeman fails with: NoMethodError: undefined method `line' for nil:NilClass


parkerfinch:
### Background

Brakeman version: 5.1.1

This issue does not occur at abe7684 but does occur at 2c086d4, so presumably the bug was introduced somewhere in there.

Rails version: 6.1.3.2
Ruby version: 2.6.6

Link to Rails application code: Code is not in a public repo.

### Issue

Brakeman is unable to complete successfully and fails with the backtrace below.
It fails on a file that includes something like:
```
{
  somekey: :someval,
  some_other_key: :some_other_val
}.values_at(*PRIMARY_KEY).join("$")
```

Brakeman completes successfully if I change the code to either:
```
{
  somekey: :someval,
  some_other_key: :some_other_val
}.values_at(*PRIMARY_KEY)
```
or
```
{
  somekey: :someval,
  some_other_key: :some_other_val
}.values_at(:literal1, :literal2).join("$")
```

so it seems like it is the combination of the splat and join that is leading to issues.

Let me know what other info I can provide!

#### Other Error

Run Brakeman with `--debug` to see the full stack trace.

I did not need to run with `--debug`, the stack trace is printed in normal operation. The `--debug` option was helpful for determining which file it was failing on, however.

Output of `bundle exec brakeman`:
```
Loading scanner...
Processing application in /Users/<project_path>
Processing gems...
[Notice] Detected Rails 6 application
Processing configuration...
[Notice] Escaping HTML by default
Parsing files...
Detecting file types...
Processing initializers...
Processing libs...ssed
bundler: failed to load command: brakeman (/Users/redacted/.gem/ruby/2.6.6/bin/brakeman)
NoMethodError: undefined method `line' for nil:NilClass
  /Users/xxx/.gem/ruby/2.6.6/bundler/gems/brakeman-2c086d45d3bc/lib/brakeman/processors/alias_processor.rb:370:in `join_item'
  /Users/xxx/.gem/ruby/2.6.6/bundler/gems/brakeman-2c086d45d3bc/lib/brakeman/processors/alias_processor.rb:324:in `block in process_array_join'
  /Users/xxx/.gem/ruby/2.6.6/bundler/gems/brakeman-2c086d45d3bc/lib/brakeman/processors/alias_processor.rb:323:in `each'
[...]

PR #1686, opened 2022-02-22
Patch to avoid issue with nil method call in alias_processor join_item

Capncavedan:
I continue to encounter the problem mentioned in https://github.com/presidentbeef/brakeman/issues/1635, even with latest 5.2.1 Brakeman.

This corrects it.


presidentbeef:
Thanks! Can you add a test?

Capncavedan:
@presidentbeef Done!  I realized I didn't need both changes, so new commit.

presidentbeef:
Thanks!

PR #1699, opened 2022-04-06
Don't exclude requested engine or lib paths if they reside in vendor

### LINKED ISSUE #1679: Prevent explicit engine/lib paths from being excluded from app tree
**Related problem description**
I was coming up with a workaround for #1677 & hit on adding the path to the gem as an engine path. Happily, it worked great locally. Unhappily, it failed on CI. We're using https://github.com/ruby/setup-ruby, and it configures Bundler to install gems into `vendor/cache`. Brakeman skips over `vendor` (good) but also ignores any additional engine or lib paths that match `vendor` (oh no).

**Description of the solution you'd like**
When I specify a lib or engine path in config or a command-line argument, have it be scanned even if it matches one of the reject patterns in `Brakeman::AppTree`.

**Alternatives you've considered**
1. **running with `--no-skip-vendor`**
  Pretty hopeless, some of the gems we're using are tricky for Brakeman to parse & I'd prefer to work on our own code instead of making PRs to third-party projects
1. **moving the gem contents to an engine**
  This is what we had previously, we're trying to separate the code so it's easier to reuse in other projects
1. **copy the code into another directory just before running Brakeman**
  The current workaround, extremely fiddly to explain to other devs how to do it before they run Brakeman locally

**Additional context**
I'm discussing how to get `ruby/setup-ruby` to not put gems in `vendor` in ruby/setup-ruby#291

jrafanie:
Fixes #1679

Currently, we find files that match various patterns such as the engine or
lib paths and exclude them if they reside in a vendor directory.

Instead, we should honor explicitly-allowed paths, even if we're excluding
vendor and they reside in a vendor directory.

Note, there are several globally EXCLUDED_PATHS such as:
/generators, test/, log/ that are still excluded even if they're in opt-in
engine/lib paths. This still seems reasonable and a test is included to
demonstrate this behavior is unchanged.

### DIFF
--- a/lib/brakeman/app_tree.rb
+++ b/lib/brakeman/app_tree.rb
@@ -205,7 +205,7 @@ module Brakeman
       paths.reject do |path|
         relative_path = path.relative
 
-        if @skip_vendor and relative_path.include? 'vendor/'
+        if @skip_vendor and relative_path.include? 'vendor/' and !in_engine_paths?(path) and !in_add_libs_paths?(path)
           true
         else
           EXCLUDED_PATHS.any? do |excluded|
@@ -213,6 +213,14 @@ module Brakeman
           end
         end
       end
+    end
+
+    def in_engine_paths?(path)
+      @engine_paths.any? { |p| path.absolute.include?(p) }
+    end
+
+    def in_add_libs_paths?(path)
+      @additional_libs_path.any? { |p| path.absolute.include?(p) }
     end
 
     def match_path files, path
--- a/test/tests/app_tree.rb
+++ b/test/tests/app_tree.rb
@@ -1,0 +1,70 @@
+require_relative '../test'
+require 'brakeman/app_tree'
+require 'fileutils'
+
+class AppTreeTests < Minitest::Test
+  def temp_dir_and_file_from_path(relative_path)
+    Dir.mktmpdir do |dir|
+      file = File.join(dir, relative_path)
+      FileUtils.mkdir_p(file)
+      FileUtils.touch(file)
+      yield dir, file
+    end
+  end
+
+  def test_ruby_file_paths
+    temp_dir_and_file_from_path("test.rb") do |dir, file|
+      at = Brakeman::AppTree.new(dir)
+      assert_equal [file], at.ruby_file_paths.collect(&:absolute)
+    end
+  end
+
+  def test_ruby_file_paths_skip_vendor_false
+    temp_dir_and_file_from_path("vendor/test.rb") do |dir, file|
+      at = Brakeman::AppTree.new(dir, :skip_vendor => false)
+      assert_equal [file], at.ruby_file_paths.collect(&:absolute)
+    end
+  end
+
+  def test_ruby_file_paths_skip_vendor_true
+    temp_dir_and_file_from_path("vendor/test.rb") do |dir, file|
+      at = Brakeman::AppTree.new(dir, :skip_vendor => true)
+      assert_equal [], at.ruby_file_paths.collect(&:absolute)
+    end
+  end
+
+  def test_ruby_file_paths_skip_vendor_true_add_libs_path
+    temp_dir_and_file_from_path("vendor/bundle/gems/gem123/lib/test.rb") do |dir, file|
+      at = Brakeman::AppTree.new(dir, :skip_vendor => true, :additional_libs_path => ["vendor/bundle/gems/gem123/lib"])
+      assert_equal [file], at.ruby_file_paths.collect(&:absolute)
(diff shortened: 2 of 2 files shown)


test/tests/app_tree.rb
```diff
@@ -0,0 +1,70 @@
...
+      FileUtils.mkdir_p(file)
+      FileUtils.touch(file)
+      yield dir, file
+    end
+  end
+
+  def test_ruby_file_paths
+    temp_dir_and_file_from_path("test.rb") do |dir, file|
+      at = Brakeman::AppTree.new(dir)
+      assert_equal [file], at.ruby_file_paths.collect(&:absolute)
```
jrafanie, line 18: These tests are testing the simplest of the public methods: `ruby_file_paths`

test/tests/app_tree.rb
```diff
@@ -0,0 +1,70 @@
...
+    temp_dir_and_file_from_path("vendor/bundle/gems/gem123/lib/test.rb") do |dir, file|
+      at = Brakeman::AppTree.new(dir, :skip_vendor => true, :additional_libs_path => ["vendor/bundle/gems/gem123/lib"])
+      assert_equal [file], at.ruby_file_paths.collect(&:absolute)
+    end
+  end
+
+  def test_ruby_file_paths_skip_vendor_true_add_engine_path
+    temp_dir_and_file_from_path("vendor/bundle/gems/gem123/test.rb") do |dir, file|
+      at = Brakeman::AppTree.new(dir, :skip_vendor => true, :engine_paths => ["vendor/bundle/gems"])
+      assert_equal [file], at.ruby_file_paths.collect(&:absolute)
```
jrafanie, line 46: This assertion fails without the code change

test/tests/app_tree.rb
```diff
@@ -0,0 +1,70 @@
...
+    temp_dir_and_file_from_path("vendor/test.rb") do |dir, file|
+      at = Brakeman::AppTree.new(dir, :skip_vendor => true)
+      assert_equal [], at.ruby_file_paths.collect(&:absolute)
+    end
+  end
+
+  def test_ruby_file_paths_skip_vendor_true_add_libs_path
+    temp_dir_and_file_from_path("vendor/bundle/gems/gem123/lib/test.rb") do |dir, file|
+      at = Brakeman::AppTree.new(dir, :skip_vendor => true, :additional_libs_path => ["vendor/bundle/gems/gem123/lib"])
+      assert_equal [file], at.ruby_file_paths.collect(&:absolute)
```
jrafanie, line 39: This assertion fails without the code change

test/tests/app_tree.rb
```diff
@@ -0,0 +1,70 @@
...
+    temp_dir_and_file_from_path("vendor/bundle/gems/gem123/test.rb") do |dir, file|
+      at = Brakeman::AppTree.new(dir, :skip_vendor => true, :engine_paths => ["vendor/bundle/gems"])
+      assert_equal [file], at.ruby_file_paths.collect(&:absolute)
+    end
+  end
+
+  def test_ruby_file_paths_skip_vendor_true_tests_in_engine_path_still_excluded
+    temp_dir_and_file_from_path("vendor/bundle/gems/gem123/test/test.rb") do |dir, file|
+      at = Brakeman::AppTree.new(dir, :skip_vendor => true, :engine_paths => ["vendor/bundler/gems"])
+      assert_equal [], at.ruby_file_paths.collect(&:absolute)
```
jrafanie, line 53: This test demonstrates that the existing global EXCLUDED_PATHS are still honored.  Even if we have tests or logs or anything else in EXCLUDED_PATHS and they reside in explicitly requested paths such as engine or lib paths, we still exclude them.

presidentbeef:
I'm really glad someone else took this on :smile:

lib/brakeman/app_tree.rb
```diff
@@ -215,6 +215,14 @@ def reject_global_excludes(paths)
       end
     end
 
+    def in_engine_paths?(path)
+      @engine_paths.any? { |p| path.absolute.include?(p) }
+    end
+
+    def in_add_libs_paths?(path)
```
jrafanie, line 222: Note, I used the shorter convention seen in the CLI options interface: https://github.com/presidentbeef/brakeman/blob/0be81f6056b8e24f05da29546379fb079bdca573/lib/brakeman/options.rb#L190

I can rename this if desired to match the instance variable's name.

jrafanie:
> I'm really glad someone else took this on 😄

@presidentbeef Is there anything else needed for this PR?
Thanks in advance!

PR #1708, opened 2022-04-27
Create docker-image.yml

kwerle:
When all is said and done, this was cookie cutter simple.  You could get more creative in setting version numbers and stuff like that, but if all you want is latest to be the latest, this does the job.

I did a test run in my account and you can see the results at https://hub.docker.com/r/kwerle/brakeman .
All I did was add a token to my settings on docker.com and drop that and my username in the secrets in the github project.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,29 @@
+name: Docker Image CI
+
+on:
+  push:
+    branches: [ main ]
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      -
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      -
+        name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: kwerle/brakeman:latest
```
presidentbeef, line 29: Should be `presidentbeef/brakeman:latest`, right?

kwerle: > Should be `presidentbeef/brakeman:latest`, right?

That ought to do!

presidentbeef:
Thank you for this. I have used #1729 instead, but I appreciate your contribution!

PR #1711, opened 2022-05-24
Fix markdown message with pipe character

Maysora:
Warning message containing codes with pipe character `|` breaks markdown table


presidentbeef:
Hi! Can you add a test?

PR #1715, opened 2022-06-08
3.2Expand Regex DoS check to include String#match and #match? coercion

### LINKED ISSUE #1714: Add Check: REDoS from match/match? coercing unsafe strings to regular expressions

Ruby's `#match` and `#match?` methods will coerce string inputs to regular expressions ([docs](https://ruby-doc.org/core-3.1.0/String.html#method-i-match)):

> Computes regexp by converting pattern (if not already a Regexp). 

Example:

```ruby
irb(main):001:0> RUBY_VERSION
=> "3.1.2"
irb(main):002:0> "haystack".match?(".*")
=> true
irb(main):003:0> "haystack".match(".*")
=> #<MatchData "haystack">
irb(main):004:0> "haystack".match? Regexp.escape(".*")
=> false
```

[This is unsafe.](https://brakemanscanner.org/docs/warning_types/denial_of_service/) Currently Brakeman checks for unescaped regex construction like [`/#{params[:regex]}/`](https://github.com/presidentbeef/brakeman/pull/454/files#diff-cbeec0c9826ea5cead02e48aafe165025665827ff9c95ef6ef621be11240fb95R74), but does not check for string coercion by `match` or `match?`.

I'd like Brakeman to add a check for `match` and `match?` called with unsafe string input. 

Thank you 🙏

bensheldon:
Closes #1714.

I added additional behavior to the existing `check_regex_dos.rb` file. I thought it seemed like the correct location because

### DIFF
--- a/lib/brakeman/checks/check_regex_dos.rb
+++ b/lib/brakeman/checks/check_regex_dos.rb
@@ -11,12 +11,20 @@ class Brakeman::CheckRegexDoS < Brakeman::BaseCheck
     ]
   }
 
-  @description = "Searches regexes including user input"
+  COERCES_STRING_TO_REGEX = [
+    :match,
+    :match?
+  ]
+
+  @description = "Searches regexes and coerced regexes including user input"
 
   #Process calls
   def run_check
     Brakeman.debug "Finding dynamic regexes"
     calls = tracker.find_call :method => [:brakeman_regex_interp]
+    COERCES_STRING_TO_REGEX.each do |coercion_method|
+      calls.concat tracker.find_call(:method => [coercion_method]).select { |call| string?(call[:target]) }
+    end
 
     Brakeman.debug "Processing dynamic regexes"
     calls.each do |call|
@@ -44,7 +52,11 @@ class Brakeman::CheckRegexDoS < Brakeman::BaseCheck
       end
 
       if match
-        message = msg(msg_input(match), " used in regular expression")
+        if result[:method] == :brakeman_regex_interp
+          message = msg(msg_input(match), " used in regular expression")
+        else
+          message = msg(msg_input(match), " used in string to regular expression coercion by ", msg_code(call.method), " method")
+        end
 
         warn :result => result,
           :warning_type => "Denial of Service",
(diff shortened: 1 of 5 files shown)


bensheldon:
oops, sorry this got submitted in a half-written state. Cat on keyboard problems 😸

lib/brakeman/checks/check_regex_dos.rb
```diff
@@ -11,12 +11,20 @@ class Brakeman::CheckRegexDoS < Brakeman::BaseCheck
...
+  ]
+
+  @description = "Searches regexes and coerced regexes including user input"
 
   #Process calls
   def run_check
     Brakeman.debug "Finding dynamic regexes"
     calls = tracker.find_call :method => [:brakeman_regex_interp]
+    COERCES_STRING_TO_REGEX.each do |coercion_method|
+      calls.concat tracker.find_call(:method => [coercion_method]).select { |call| string?(call[:target]) }
```
presidentbeef, line 26: Could simplify this to

```ruby
tracker.find_call(:methods => CEORCES_STRING_TO_REGEX).select { ... }
```

presidentbeef (changes requested):

bensheldon:
@presidentbeef I updated the code based on your feedback. Thank you! 🙏🏻 

I also made one behavior change: I added a method called `string_or_modified_string?` that recurses down the target to see if the initial call was a String. I added this because I found making a trivial modification like `downcase` would not trigger the check (e.g. `"haystack".downcase.match(params[:unsafe])`). I couldn't find an existing recursive method, but maybe I overlooked one.

presidentbeef:
In real code, would we expect to see string literals like the test cases? :thinking:

bensheldon:
> In real code, would we expect to see string literals like the test cases? 🤔

This was the code we discovered (lightly anonymized):

```ruby
result = @active_record_resource.a_hash_value
  .keys
  .select { |key| key.downcase.match?(unsafe_string_query) }
  .sort
```

I'd love advice for better ways to identify `[String]#match?([String])`.

Issue #1545, opened 2021-01-14
Report on: allow_forgery_protection = false


forced-request:
**Is your feature request related to a problem? Please describe.**

@presidentbeef and I have talked about this a bit already, but the gist is that I recommend adding a rule to check for any changes to `allow_forgery_protection` via global configuration. 

In Rails 6 users can explicitly opt-out of CSRF protections at the config level by setting `config.action_controller.allow_forgery_protection = false`, as seen in the example below.

When this situation occurs Brakeman fails to warn about it, resulting in the application being completely vulnerable to CSRF.

```ruby
require_relative 'boot'

require 'rails/all'

# Require the gems listed in Gemfile, including any gems
# you've limited to :test, :development, or :production.
Bundler.require(*Rails.groups)

module Test
  class Application < Rails::Application
    # Initialize configuration defaults for originally generated Rails version.
    config.load_defaults 6.0

    config.action_controller.allow_forgery_protection = false

    config.active_record.schema_format = :sql
  end
end
```

**Describe the solution you'd like**
* Brakeman should warn if `allow_forgery_protection` is disabled.
* Brakeman should warn (with lower severity) if `per_form_csrf_tokens` is disabled

**Describe alternatives you've considered**
N/A

**Additional context**

jamgregory:
I think that sounds like a sensible idea to me. I was also wondering if you could bundle the following check in as well?

```
skip_before_action :verify_authenticity_token
```

I think it falls into the same category - it's explicitly stopping the check of the CSRF token, and should probably be warned about by Brakeman.

jamgregory:
Reading more into this, I wonder if it's also worth flagging up cases where `protect_from_forgery` [could be turned off](https://apidock.com/rails/v6.1.3.1/ActionController/RequestForgeryProtection/ClassMethods/protect_from_forgery), e.g.:

```
protect_from_forgery except: :index
protect_from_forgery only: :show
protect_from_forgery if: -> { ... }
protect_from_forgery unless: -> { ... }
```

PR #1719, opened 2022-07-06
Fix `load_rails_defaults` overwriting settings in the Rails application

jamgregory:
While doing some work to implement some additional checks for #1545, I discovered that while Brakeman loads the Rails configuration as expected, if any defaults are set, these are actually overwritten and so Brakeman won't report issues where those defaults are different inside the application.

This appears to be at odds to how Rails works with defaults, where they are (usually) overwritten by the environment configuration.

### DIFF
--- a/lib/brakeman/processors/lib/rails3_config_processor.rb
+++ b/lib/brakeman/processors/lib/rails3_config_processor.rb
@@ -86,7 +86,7 @@ class Brakeman::Rails3ConfigProcessor < Brakeman::BasicProcessor
       end
     elsif include_rails_config? exp
       options_path = get_rails_config exp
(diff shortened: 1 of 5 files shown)


presidentbeef (changes requested):
Changes look okay but not keen on adding an entire new test application.

Can this be tested by updating one of the existing apps?

jamgregory:
> Changes look okay but not keen on adding an entire new test application.
> 
> Can this be tested by updating one of the existing apps?

Ah, of course - no worries at all. I've tweaked the Rails 5.2 app and then used the rescanner to make the changes I was testing for. I hope that is OK?

presidentbeef:
Thanks!

PR #1729, opened 2022-08-11
Add GitHub Action to publish docker image for new releases

### LINKED ISSUE #1680: `latest` brakeman image in docker hub is older than `5.2.1`
As a follow-up to #1510, I thought I'd raise the same issue. Any chance we could get the brakeman image updated on Dockerhub? Thank you!

ShockwaveNN:
Resolve #1680

This will build both `amd64` and `arm64` image

Maintainers should add secrets for `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` in GitHub repo settings

Not 100% that it's a working actions file, since it will only trigger on new tags creation, but it's based on some of my other repo with minor differences

Maintainers can test it by creating some temp tag (after this merged) like `test-tag-please-ignore` and check that docker images are built and pushed to docker hub (this will override `latest` image through, but it's already very outdated, so don't think there is any harm)

### DIFF
--- a/.github/workflows/docker-hub-push.yml
+++ b/.github/workflows/docker-hub-push.yml
@@ -1,0 +1,31 @@
+name: docker-hub-push
+on:
+  push:
+    tags:
+      - '*'
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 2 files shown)


presidentbeef:
Cool, thanks for this! I will check it out soon.

presidentbeef:
I have confirmed this works - thanks!

ShockwaveNN:
@presidentbeef Glad to hear that )

PR #1737, opened 2022-10-24
Add new check for weak RSA keys and padding modes

### LINKED ISSUE #1736: Add checks for weak public-key cryptography
Similar to [check_weak_hash.rb](https://github.com/presidentbeef/brakeman/blob/main/lib/brakeman/checks/check_weak_hash.rb) and inspired by https://github.com/ruby/openssl/pull/546 - it is quite easy to use RSA and other asymmetric algorithms insecurely. Some ideas that can be checked for:

- Small key sizes (eg RSA < 2048 bits = [CWE-326](https://cwe.mitre.org/data/definitions/326.html)) when generating keys
- Insecure padding modes (eg RSA PKCS#1 v1.5 = [CWE-780](https://cwe.mitre.org/data/definitions/780.html)) when doing operations 

Flag the code like this [blog post](https://shopify.engineering/building-flex-comp) which specifically showcases CWE-780:
```ruby
def encrypt(payload)
  public_key = OpenSSL::PKey::RSA.new("grab the public 4096 bit key")
  Base64.encode64(public_key.public_encrypt(payload.to_json)
end
```

`OpenSSL::PKey::RSA#public_encrypt` defaults to insecure PKCS#1 v1.5 padding. On modern OpenSSL gems (>= 3.0) this is [a wrapper](https://github.com/ruby/openssl/blob/fd5eaa6dfc9e70875505cbe89a854223e7922776/lib/openssl/pkey.rb#L429-L449) for `OpenSSL::PKey::PKey#encrypt`, so effectively the following code is called:
```ruby
public_key.encrypt(payload.to_json, "rsa_padding_mode" => "pkcs1")
```

Both types of calls should be flagged.

**Describe alternatives you've considered**
N/A

presidentbeef:
Fixes #1736

Warns about these patterns:

* `OpenSSL::PKey::RSA.new` or `OpenSSL::PKey::RSA.generate` with key size < 2048 bits
* `OpenSSL::PKey::RSA#public_encrypt`, `OpenSSL::PKey::RSA#public_decrypt`, `OpenSSL::PKey::RSA#private_encrypt`, `OpenSSL::PKey::RSA#private_decrypt` with no padding mode specified (in which case it's PKCS1), with `OpenSSL::PKey::RSA::PKCS1`, or `OpenSSL::PKey::RSA::NO_PADDING`


bdewater:
Thank you for working on this! Some feedback:

https://cwe.mitre.org/data/definitions/780.html is quite strong in calling anything other than OAEP padding secure, should we follow that recommendation? I hope nobody picks for example `SSLV23_PADDING` on purpose, but then again who knows if people don't know why `PKCS1_PADDING` is insecure.

The PR currently does not seem to detect calls using the generic PKey interface [introduced](https://github.com/ruby/openssl/pull/374) in OpenSSL gem 3.0:
```ruby
weak_rsa = OpenSSL::PKey.generate_key("RSA", rsa_keygen_bits: 1024)
weak_encrypted = weak_rsa.encrypt("data", rsa_padding_mode: "pkcs1")
weak_signature_digest = weak_rsa.sign("SHA256", "data", rsa_padding_mode: "pkcs1")
weak_signature_raw = weak_rsa.sign_raw(nil, "data", rsa_padding_mode: "pkcs1")
```

presidentbeef:
Makes sense, I can add those in as well.

presidentbeef:
Now warns on key size for `OpenSSL::PKey.generate_key`.

Warns on any padding mode other than `OAEP`.

Added padding checks for `encrypt`, `decrypt`, `sign`, `verify`, `sign_raw`, `verify_raw`.

PR #1749, opened 2022-12-06
Prevent redirects using url_from being marked as unsafe

lsylvester:
Rails has introduced a methd [url_from](https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-url_from) for guarding against unsafe redirects.

It is idiomatically used as 

```
redirect_to url_from(params[:redirect_url]) || root_url
```

This PR introduces two changes to the Redirect check to fully support this.

First - any redirect using `url_from` (such as `redirect_to url_from(params[:r])`) is now considered safe.
Second - adding fallback to the url (such as `redirect_to params[:r] || "/"`) no longer fails to produce a warning

### DIFF
--- a/lib/brakeman/checks/check_redirect.rb
+++ b/lib/brakeman/checks/check_redirect.rb
@@ -45,7 +45,7 @@ class Brakeman::CheckRedirect < Brakeman::BaseCheck
         not explicit_host?(opt) and
         not slice_call?(opt) and
         not safe_permit?(opt) and
-        res = include_user_input?(call)
+        res = include_user_input?(opt)
 
       if res.type == :immediate
         confidence = :high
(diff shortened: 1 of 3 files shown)


lib/brakeman/checks/check_redirect.rb
```diff
@@ -68,42 +68,44 @@ def process_result result
...
-        return Match.new(immediate, arg)
-      elsif request_value? arg.target
-        return Match.new(immediate, arg.target)
-      elsif arg.method == :url_for and include_user_input? arg
-        return Match.new(immediate, arg)
+    elsif call? opt
+      if request_value? opt
+        return Match.new(immediate, opt)
+      elsif request_value? opt.target
+        return Match.new(immediate, opt.target)
```
lsylvester, line 92: From the local coverage report - I think that this is the line that is missing the coverage in the diff. 

I am not sure what would match this condition without also matching `request_value? opt`.

presidentbeef:
Ah this is really cool.

lib/brakeman/checks/check_redirect.rb
presidentbeef: I think you are right.

presidentbeef:
Since Rails 7.0 now [automatically protects from open redirects](https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_to), it looks like Brakeman needs to check for `raise_on_open_redirects` and potentially warn about `allow_other_host: true` :thinking:

PR #1759, opened 2023-02-10
Add ruby 3.2 to CI

stoivo:
Something fails on when I use ruby 3.2 in production. I think this should work but it is allways good to test have it in ci too.

### DIFF
--- a/test/tests/rails7.rb
+++ b/test/tests/rails7.rb
@@ -5,7 +5,10 @@ class Rails7Tests < Minitest::Test
   include BrakemanTester::CheckExpected
 
   def report
-    @@report ||= BrakemanTester.run_scan "rails7", "Rails 7", :run_all_checks => true
+    @@report ||=
+      Date.stub :today, Date.parse('2023-02-10') do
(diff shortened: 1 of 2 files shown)


stoivo:
Sorry if you think I changed to many things in one MR. It is less work on one MR when it soo related.

test/tests/rails7.rb
```diff
@@ -13,10 +13,16 @@ def expected
       :controller => 0,
       :model => 0,
       :template => 0,
-      :warning => 22
+      :warning => 23
     }
   end
 
+  def test_ruby_2_7_eol
+    assert_warning check_name: "EOLRuby",
```
presidentbeef, line 21: This will break again on 2023-03-31 (or maybe sooner).

Can you wrap it in code like this? (This is what's done [here](https://github.com/presidentbeef/brakeman/blob/main/test/tests/cves.rb#L351))

```ruby
Date.stub :today, Date.parse('2022-02-10') do
  assert_warning ...
end
```

stoivo: It is updated. I wraped the scann method

presidentbeef:
Thanks!

Issue #1139, opened 2017-12-08
Unscoped find on self-referential relationship


CoryMcDonald:
### Background

Brakeman version: 4.0.1
Rails version: 5.1.4
Ruby version: 2.4.1

Link to Rails application code: https://github.com/CoryMcDonald/ExampleRailsApplication/

### Issue

When setting up a self-referential association I found this interesting case where if I call `.find` in a controller than I get an "unscoped find" warning. 

#### False Positive

*Full* warning from Brakeman:
```
== Warnings ==

Confidence: Weak
Category: Unscoped Find
Check: UnscopedFind
Message: Unscoped call to User#find
Code: User.find(params[:id])
File: app/controllers/user_controller.rb
Line: 3
```
Relevant code:

```ruby
class User < ApplicationRecord
  has_one  :match, class_name: 'User', foreign_key: :matched_user
  belongs_to :matched_user, class_name: 'User', optional: true
end
```

_Why might this be a false positive?_

1. The `belongs_to` association is listed as `optional: true`. (The syntax differs between Rails 4/5) 
2. Because the relationship is self-referential I don't believe it is reasonable to expect `find` actions to be vulnerable as you'd be hard-pressed to be able to scope the model without using `find`.

Perhaps a potential solution would be to simply check to see if the relationship is optional and not report that if that's the case.

PR #1764, opened 2023-03-03
Fixed detection of UnscopedFind if optional: true is defined in the model and is not self-reference related

NerdyBoyCool:
ref https://github.com/presidentbeef/brakeman/issues/1139
ref https://github.com/presidentbeef/brakeman/pull/1153

The change in https://github.com/presidentbeef/brakeman/pull/1153 avoids checking UnscopedFind for model classes with optional: true settings, but I think UnscopedFind warning should be issued for the following cases.

The current code only checks if optional: true is defined in the model class, so the check for the Group model is skipped.

```rb
class Group < ApplicationRecord
  belongs_to :user, optional: true
end
```

```rb
class GroupsController < ApplicationController
  def show
    @group = Group.find(params[:id])
  end
end
```

This PR solves this problem by also checking self-reference associations.
We have also added implementation and testing for the issues mentioned in https://github.com/presidentbeef/brakeman/issues/1139 so that they are not affected.

### DIFF
--- a/lib/brakeman/checks/check_unscoped_find.rb
+++ b/lib/brakeman/checks/check_unscoped_find.rb
@@ -11,13 +11,13 @@ class Brakeman::CheckUnscopedFind < Brakeman::BaseCheck
 
     associated_model_names = active_record_models.keys.select do |name|
       if belongs_to = active_record_models[name].associations[:belongs_to]
-        not optional_belongs_to? belongs_to
+        not self_referential_association? belongs_to, name
       else
         false
       end
     end
 
-    calls = tracker.find_call :method => [:find, :find_by_id, :find_by_id!],
+    calls = tracker.find_call :method => [:find, :find_by_id, :find_by_id!, :find_by, :find_by!],
                               :targets => associated_model_names
 
     calls.each do |call|
@@ -52,12 +52,16 @@ class Brakeman::CheckUnscopedFind < Brakeman::BaseCheck
       :cwe_id => [285]
   end
 
-  def optional_belongs_to? exp
+  def self_referential_association? exp, name
     return false unless exp.is_a? Array
 
     exp.each do |e|
       if hash? e and true? hash_access(e, :optional)
-        return true
+        hash_iterate e do |key, value|
+          if key[1] == :class_name and name == value[1].to_sym
+            return true
+          end
+        end
       end
     end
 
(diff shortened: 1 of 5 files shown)


NerdyBoyCool:
@presidentbeef 
Can you give me your opinion on this change?
I would appreciate any advice and opinions.

presidentbeef:
I'm not sure I understand the intent here. If the relationship is optional, wouldn't you have legitimate cases where the application can't look up a model via the relationship..? In other words, is this change just as likely to increase false positives as it is to decrease false negatives?

NerdyBoyCool:
The first example I gave was a little confusing, so I will give an example that is similar to the problem we encountered.

```rb
class User < ApplicationRecord
end

class Organization < ApplicationRecord
  has_many :members, dependent: :destroy
  has_many :users, dependent: :destroy
end

class Member < ApplicationRecord
  belongs_to :user
  belongs_to :organization

  has_many :articles, dependent: :destroy
end

class Article < ApplicationRecord
  belongs_to :author, class_name: "User", foreign_key: "author_id", optional: true
  belongs_to :organization
end

# Bad
artciles = Article.find(params[:id])

# Good
organization = Organization.first
articles = organization.articles.find(params[:id])
```

I expected brakeman to detect that the Article was not being retrieved via an Organization instance, but it did not because the Article model already defines optional: true for the User.
As you say, this type of problem could increase the number of false positives, but I thought the problem would be greater if the code that calls find directly against Organization was skipped.

lib/brakeman/checks/check_unscoped_find.rb
```diff
@@ -52,12 +52,16 @@ def process_result result
       :cwe_id => [285]
   end
 
-  def optional_belongs_to? exp
+  def self_referential_association? exp, name
     return false unless exp.is_a? Array
 
     exp.each do |e|
```
presidentbeef, line 58: I think what we actually want is to check if there is any `belongs_to` relationship that is _not_ optional?

Is the self-referential piece actually that important?

lib/brakeman/checks/check_unscoped_find.rb
```diff
@@ -52,12 +52,16 @@ def process_result result
...
       if hash? e and true? hash_access(e, :optional)
-        return true
+        hash_iterate e do |key, value|
```
presidentbeef, line 60: Is `class_name` going to show up more than once as a key? Couldn't this also use `hash_access(:class_name)` instead of iterating?

willnet:
@NerdyBoyCool I’m interested in this PR and can address the remaining items. Would it be alright if I took it over?

NerdyBoyCool:
@willnet 
Hi. Of course!!
I’m not able to spend time on this, so please take it over. Thanks!!!

PR #1771, opened 2023-03-24
Drop older Rubies

presidentbeef:
- Raise required Ruby to 3.0.
- Drop parsing of Ruby 1.8 and 1.9

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 jobs:
   default: &default
     docker:
-      - image: cimg/ruby:2.7
+      - image: cimg/ruby:3.0
     steps:
       - checkout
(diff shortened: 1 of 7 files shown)


armahillo:
In the future, can requiring a new ruby version cause a bigger bump in the semver versioning? (minor bump, at a minimum)

We had '~> 5.4' and weren't yet ready to update ruby 3 (still resolving some other compatibility issues).

presidentbeef:
@armahillo ❓ ❓ 

I'm not sure what you mean. This change was released in Brakeman 6.0, a major version update. If you are on 5.x then the latest is [5.4.1](https://rubygems.org/gems/brakeman/versions/5.4.1) which requires Ruby 2.5+

PR #1776, opened 2023-04-20
Fix default cross-site request forgery setting when defaults are not loaded

### LINKED ISSUE #1773: False Negative: warning on CSRF in Rails 5.2+ with defaults
### Background

Brakeman version: ? 5.4.0
Rails version: ? 6.0
Ruby version: ? 2.7

Link to Rails application code:  closed-source

#### False Positive

When:

- `load_defaults` is NOT present
- `default_protect_from_forgery` is NOT configured anywhere

*Full* warning from Brakeman: 

No warning, but there should be a security warning generated by `checks/check_forgery_setting.rb`.

Relevant code:

```ruby
# these lines are MISSING from your config
# load_defaults 5.2 
# Rails.application.config.action_controller.default_protect_from_forgery = nil
```

_Why might this be a false negative?_

Because by default a Rails 6 app does NOT have protect for forgery enabled unless `load_defaults 5.2` or higher is used...

```
> Rails.application.config.action_controller.default_protect_from_forgery
=> nil
> ApplicationController.default_protect_from_forgery
=> false
```

The logic for this check assumes that defaults are tied to the Rails version number only... but this is untrue, they are based on the actual underlying defaults of Rails (often nil) PLUS the `load_defaults` one is using - and it's entirely possible (though perhaps ill-advised) not to use load_defaults at all.

presidentbeef:
Follow up from #1530, adding a test.

Should fix #1773

### DIFF
--- a/lib/brakeman/tracker/config.rb
+++ b/lib/brakeman/tracker/config.rb
@@ -20,9 +20,7 @@ module Brakeman
 
     def default_protect_from_forgery?
       if version_between? "5.2.0.beta1", "9.9.9"
-        if @rails.dig(:action_controller, :default_protect_from_forgery) == Sexp.new(:false)
-          return false
-        else
+        if @rails.dig(:action_controller, :default_protect_from_forgery) == Sexp.new(:true)
           return true
         end
       end
--- a/test/tests/rails52_csrf.rb
+++ b/test/tests/rails52_csrf.rb
@@ -1,0 +1,38 @@
+require_relative '../test'
+require 'brakeman/rescanner'
+
+class Rails52CSRFTest < Minitest::Test
+  include BrakemanTester::RescanTestHelper
+  include BrakemanTester::FindWarning
+
+  def report
+    @report
+  end
+
+  def test_csrf_with_no_load_defaults
+    tracker = nil
+
+    # Terribly abusing the rescan functionality here.
+    # Actually don't want the rescan, just want to run a regular scan
+    # because we don't have the capability to rescan with on config changes
+    # like this and I don't feel like building it right now.
+    before_rescan_of ['config/application.rb'], 'rails5.2' do |app_dir|
+      replace 'config/application.rb', 'config.load_defaults 5.2', ''
+      tracker = Brakeman.run(app_path: app_dir, parallel_checks: false)
+    end
+
+    @report = tracker.report.to_hash
+
+    assert_warning check_name: "ForgerySetting",
+      type: :controller,
+      warning_code: 7,
+      fingerprint: "6f5239fb87c64764d0c209014deb5cf504c2c10ee424bd33590f0a4f22e01d8f",
+      warning_type: "Cross-Site Request Forgery",
+      line: 1,
+      message: /^`protect_from_forgery`\ should\ be\ called\ /,
+      confidence: 0,
+      relative_path: "app/controllers/application_controller.rb",
+      code: nil,
+      user_input: nil
+  end
+end


joshgoebel:
Yes, I think this inversion on the conditional would fix it!  Thanks!

MaksJS:
I've got a regression with this one on Rails 7. `Rails.configuration.action_controller.default_protect_from_forgery` is true and I get the `Cross-Site Request Forgery` warning. Maybe, at some point, the value of `default_protect_from_forgery` is not passed to the tracker config when not inlined in the configuration file (true is the default value).

joshgoebel:
Where else are you setting the configuration?  I guess I'd expect it to run part of the init to get the values - but does it only do static analysis?

presidentbeef:
For security reasons, Brakeman does not run any of the code it analyzes.

@MaksJS Where is `default_protect_from_forgery` set in your application? Is it in an initializer?

MaksJS:
@joshgoebel @presidentbeef Nowhere, it's `true` by default so I don't set it again in the config

joshgoebel:
> t's true by default 

So you're using `load_defaults`?

MaksJS:
@joshgoebel Yes

presidentbeef:
@MaksJS I cannot reproduce your issue.

This is what the [Rails 7 test app](https://github.com/presidentbeef/brakeman/blob/main/test/apps/rails7/config/application.rb#L12) looks like:

```ruby
module Rails7
  class Application < Rails::Application
    # Initialize configuration defaults for originally generated Rails version.
    config.load_defaults 7.0
``` 

It's very hard to guess at the issue. Please share the warning you are getting what your code looks like. Even better, share an example Rails application that demonstrates the issue. Thanks!

MaksJS:
@presidentbeef Ok I was able to reproduce it. You're right, with a brand new Rails 7 application it works well.
But on my case, I'm developing a Rails Engine with a dummy application, which have this configuration by default:

```ruby
module Dummy
  class Application < Rails::Application
    config.load_defaults Rails::VERSION::STRING.to_f
  end
```

This configuration is generated by this template: https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/config/application.rb.tt#L15

In this case, I get the `Cross-Site Request Forgery` warning with Brakeman 6.0.0.

presidentbeef:
Ah. In that case Brakeman cannot know for which version of Rails to load the defaults. What version would you expect it to use? Is there a `Gemfile.lock` with a specific version?

MaksJS:
Well, `Rails::VERSION::STRING.to_f` would evaluate to `7.0` here. You can also get the version number with `Rails.configuration.loaded_config_version`, so maybe it can be available in `tracker.config.rails.dig(:loaded_config_version)`. Or maybe not if the app config is never evaluated, which seems to be the case if I understand correctly this method https://github.com/presidentbeef/brakeman/blob/6af53c63feb909d19bab970aedb3b0c583073eb6/lib/brakeman/tracker/config.rb#L75

But yes there is an entry in `Gemfile.lock`: `rails (~> 7.0.5)`.

And I correctly get the version number in the report:

```
[Notice] Detected Rails 7 application
....
Rails Version: 7.0.5
Brakeman Version: 6.0.0
....
== Warnings ==

Confidence: High
Category: Cross-Site Request Forgery
....
```

joevin-slq-docto:
Hello,
I'm using Rails 7 with `config.load_defaults 6.0 `inside `application.rb`.
Rails console confirms me that `Rails.configuration.action_controller.default_protect_from_forgery` is set to `true` but I get the Cross-Site Request Forgery warning.

PR #1789, opened 2023-08-10
Add PG::Connection.escape_string as authorized sanitization function

joevin-slq-docto:
Hello,

In the same way as [ActiveRecord::Base.sanitize_sql()](https://api.rubyonrails.org/v7.0.6/classes/ActiveRecord/Sanitization/ClassMethods.html#method-i-sanitize_sql), I suggest adding [PG::Connection.escape_string()](https://deveiate.org/code/pg/PG/Connection.html#method-c-escape_string-doc) as an authorized function to sanitize sql statments.

Using this will never be as good as using prepared statements, but this addition, if deemed relevant, would limit false positives.

Thanks!
Nb: [escape](https://deveiate.org/code/pg/PG/Connection.html#method-i-escape) is an alias for [escape_string](https://deveiate.org/code/pg/PG/Connection.html#method-i-escape_string).


presidentbeef:
Cool, thanks!

Issue #1786, opened 2023-07-14
Unscoped find not alerted for `find_by!`


rbclark:
### Background

Brakeman version: 6.0.0
Rails version: 7.0.5.1
Ruby version: 3.2.2

Link to Rails application code:

```
Model.find_by!(public_find: params[:id]
Model.find_by!(id: params[:id])
```

### Issue

 - Brakeman has been updated to include [`find_by`](https://github.com/presidentbeef/brakeman/issues/1626) but it does not include `find_by!`.
 - Brakeman only checks for find_by on `id` and doesn't look for any other attributes. In our project the attribute we use `public_id` instead of `id` as the publicly facing ID of our models. This means we don't get any warnings for unscoped find for most of our models.

presidentbeef:
A year later...

Happy to add `find_by!`, but `public_id` is not a standard column as far as I've found.

PR #1797, opened 2023-09-17
Check Unscoped AR queries with find_by/find_by!

thukim:
Fixes 2 issues described in #1786

> Brakeman has been updated to include find_by but it does not include find_by!.
Brakeman only checks for find_by on id and doesn't look for any other attributes. In our project the attribute we use public_id instead of id as the publicly facing ID of our models. This means we don't get any warnings for unscoped find for most of our models.

### DIFF
--- a/lib/brakeman/checks/check_unscoped_find.rb
+++ b/lib/brakeman/checks/check_unscoped_find.rb
@@ -24,12 +24,9 @@ class Brakeman::CheckUnscopedFind < Brakeman::BaseCheck
       process_result call
     end
 
-    tracker.find_call(:method => :find_by, :targets => associated_model_names).each do |result|
+    tracker.find_call(:method => [:find_by, :find_by!], :targets => associated_model_names).each do |result|
       arg = result[:call].first_arg
-
-      if hash? arg and hash_access(arg, :id)
-        process_result result
-      end
+      process_result(result) if hash?(arg)
     end
   end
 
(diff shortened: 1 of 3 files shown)


thukim:
Hi @presidentbeef ,

I think this repository is having flaky tests (screenshot attached below)

Screenshot 2023-09-19 at 18 30 06

Ref: https://app.circleci.com/pipelines/github/presidentbeef/brakeman?branch=pull%2F1797

The same set of failed tests happen for `test-3-2` first and then `test-3-1` and it seems to be related to the file `lib/sweet_lib.rb`
Screenshot 2023-09-19 at 18 32 15

presidentbeef:
Yes, that is a flaky test. Sorry for the trouble. Now I'm finally motivated to fix it - fix incoming shortly, I hope.

presidentbeef:
Should be fixed with #1798. Please rebase on main and try again.

thukim:
Thanks, @presidentbeef ! After merging the latest changes from the main branch to this branch, all the tests are green 👍 
This PR is also ready to be reviewed.

presidentbeef:
Warning on all uses of `find_by`/`find_by!` is a pretty major change.

The intent of the original code was to warn about equivalents to `find` or `find_by_id`, and the overall intent of the check is to warn about insecure direct object references - i.e. uses of the row id directly.

This check already has pretty high false positive rate (which is why it is off by default), so I'm pretty hesitant to make it worse.

thukim:
I see, thanks @presidentbeef! I will close this PR then

PR #1804, opened 2023-10-27
Add multiple file for ignore options

baldarn:
Hi!

I needed the `-i` parameter to accept multiple files, because we are planning to have multiple list of ignore files based on priority to solve.

This pr doesn't change the interface, but instead add the possiblity to pass multiple paths of ignore config to the `-i` parameter

What do you think? Can this be useful?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ test/coverage/
 .bundle
 bundle
 *.gem
+test-report.json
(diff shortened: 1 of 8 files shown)


presidentbeef:
Seems a little complicated for the benefits. I'm actually not sure I understand the benefits, actually.

Editing/saving multiple files seems pretty tricky. Would you track which ignored warning came from which file..? Where would you write newly ignored warnings..?

baldarn:
You are right... we went back to the mono file, adding a special note for wont fix issues.

the goal was having a list of WONTFIX and a list of task to schedule for maintenance, but this is doable with only a file.

thanks for you time!

closing this ;)

PR #1805, opened 2023-11-08
fix: avoid detecting phlex components as dynamic render paths

ElMassimo:
### Description 📖 

This pull request adds support for [`phlex`](https://github.com/phlex-ruby/phlex), preventing brakeman from detecting usages of Phlex components as occurrences of Dynamic Render Paths.

### Background 📜 

This problem is similar to what used to happen in the past for `view_components`:
- #1529
- #1578


presidentbeef:
Can you add a test or two?

PR #1806, opened 2023-11-20
Add Trunk Check instructions

joshmarinacci:
Hi. I'm an engineer at Trunk, which uses and automates Brakeman. This PR is just to add instructions for how to enable Brakeman if you are using our free CLI. Please let me know if this isn't appropriate. Thanks!


presidentbeef:
Thanks, but I don't think it's appropriate to market your product like this. Good luck.

PR #1821, opened 2024-01-25
fix: avoid detecting 'ViewComponentContrib::Base' as dynamic render paths

vividmuimui:
# Description 📖
This pull request adds support for [view_component-contrib gem](https://github.com/palkan/view_component-contrib).
`view_component-contrib` is a support library for [ViewComponent](https://github.com/github/view_component), which defines ViewComponent using the `class SomeComponent < ViewComponentContrib::Base` instead of `class SomeComponent < ViewComponent::Base`.
(`ViewComponentContrib::Base` is a class that extends `ViewComponent::Base` (see: https://github.com/palkan/view_component-contrib/blob/master/lib/ view_component_contrib/base.rb))

This PR prevents brakeman from detecting a Component using ViewCompontontrib as an occurrence of Dynamic Render Paths.

# Background 📜

This problem is similar to what used to happen in the past for view_components:

https://github.com/presidentbeef/brakeman/issues/1529
https://github.com/presidentbeef/brakeman/pull/1578
https://github.com/presidentbeef/brakeman/pull/1805

### DIFF
--- a/lib/brakeman/checks/check_render.rb
+++ b/lib/brakeman/checks/check_render.rb
@@ -108,6 +108,11 @@ class Brakeman::CheckRender < Brakeman::BaseCheck
   def known_renderable_class? class_name
     klass = tracker.find_class(class_name)
     return false if klass.nil?
-    klass.ancestor?(:"ViewComponent::Base") || klass.ancestor?(:"Phlex::HTML")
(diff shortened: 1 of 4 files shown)


presidentbeef:
Thanks!

PR #1828, opened 2024-02-02
Add support for directory symlinks

### LINKED ISSUE #1817: Brakeman does not follow directory symlinks
### Background

Brakeman version:  6.1
Rails version: 7.1.3
Ruby version: 3.3.0

Link to Rails application code: 

I've created a minimal sample project here https://github.com/lubert/sample-rails-brakeman to demonstrate the bug

### Issue
* `brakeman` doesn't seem to follow directory symlinks. File symlinks are OK.
* In the sample project I've provided,`app/dependencies` has 2 directories, `bad_dep_symlink` is a symlink to `../../bad_dependency`, `bad_dep_no_symlink` is a copy of the same directory.
* Running `brakeman app` from the root only flags warnings with `app/dependencies/bad_dep_no_symlink`, whereas I expect `bad_dep_symlink` to also have warnings

lubert:
Fixes #1817 

* Update glob_files to follow directory symlinks
* Extract logic of `root_search_pattern` into `search_pattern` in order to apply it to other directories
* Add test for relative and absolute directory symlinks

### DIFF
--- a/lib/brakeman/app_tree.rb
+++ b/lib/brakeman/app_tree.rb
@@ -246,13 +246,16 @@ module Brakeman
 
     def root_search_pattern
       return @root_search_pattern if @root_search_pattern
+      @root_search_pattern = search_pattern(@root)
+    end
 
+    def search_pattern(root_dir)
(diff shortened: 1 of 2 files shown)


presidentbeef:
Thanks for this, I'll take a closer look.

PR #1835, opened 2024-03-24
Add test-report.json to .gitignore

tobyhs:
It is generated from running tests and can be ignored by git.


presidentbeef:
Thanks for pointing this out! Will fix the root problem in #1836

PR #1837, opened 2024-04-07
Add option to exclude "updated" in brakeman.ignore

tobyhs:
The "updated" entry in brakeman.ignore is prone to merge conflicts, so I'm adding an option that allows users to omit it.

### DIFF
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -193,6 +193,10 @@ To create and manage this file, use:
 
     brakeman -I
 
+To exclude the "updated" entry in brakeman.ignore, use:
+
+    brakeman --exclude-updated-in-ignore-file
+
 To ignore possible XSS from model attributes:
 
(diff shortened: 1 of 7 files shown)


presidentbeef:
I'm actually tempted to drop the timestamp altogether instead of adding a very specific option.

tobyhs:
> I'm actually tempted to drop the timestamp altogether instead of adding a very specific option.

That seems reasonable. I only implemented an option because removing the timestamp altogether breaks compatibility, but I'm not sure if anyone really relies on the updated entry in brakeman.ignore.

tobyhs:
I opened a pull request to drop the updated entry altogether at https://github.com/presidentbeef/brakeman/pull/1860

tobyhs:
Closing because https://github.com/presidentbeef/brakeman/pull/1860 was merged

Issue #1841, opened 2024-04-30
brakeman still references haml 4 - which is a bit long in the tooth (Haml::Filter::Coffee class vs. module)


kwerle:
In short, brakeman declares Haml::Filter::Coffee to be a module and haml declares it to be a class.

### Background

Brakeman version: 
```
$ brakeman --version
brakeman 6.1.2

```
Rails version: 
```
$ rails --version
Rails 7.0.8.1

```
Ruby version: 
```
$ ruby --version
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) +YJIT [aarch64-linux]

```
Link to Rails application code: Private

### Issue

```
Error: Coffee is not a module /.../.ruby/gems/haml-6.3.0/lib/haml/filters/coffee.rb:4: previous definition of Coffee was here While processing /.../....html.haml
Location: /.../.ruby/gems/brakeman-min-6.1.2/lib/brakeman/parsers/haml_embedded.rb:29:in `<main>'

```
HAML says:
```
$ cat /.../.ruby/gems/haml-6.3.0/lib/haml/filters/coffee.rb
# frozen_string_literal: true
module Haml
  class Filters
    class Coffee < TiltBase
      def compile(node)
        require 'tilt/coffee' if explicit_require?('coffee')
        temple = [:multi]
        temple << [:static, "<script>\n"]
        temple << compile_with_tilt(node, 'coffee', indent_width: 2)
        temple << [:static, "</script>"]
        temple
      end
    end

    CoffeeScript = Coffee
  end
end

```
Brakeman says:
```
$ cat /.../.ruby/gems/brakeman-min-6.1.2/lib/brakeman/parsers/haml_embedded.rb
module Brakeman
  module FakeHamlFilter
    # Copied from Haml 4 - force delayed compilation
    def compile(compiler, text)
      filter = self
      compiler.instance_eval do
        text = unescape_interpolation(text).gsub(/(\\+)n/) do |s|
          escapes = $1.size
          next s if escapes % 2 == 0
          ("\\" * (escapes - 1)) + "\n"
        end
        # We need to add a newline at the beginning to get the
        # filter lines to line up (since the Haml filter contains
        # a line that doesn't show up in the source, namely the
        # filter name). Then we need to escape the trailing
        # newline so that the whole filter block doesn't take up
        # too many.
        text = "\n" + text.sub(/\n"\Z/, "\\n\"")
        push_script <<RUBY.rstrip, :escape_html => false
[...]

PR #1842, opened 2024-04-30
https://github.com/presidentbeef/brakeman/issues/1841

kwerle:
brakeman still references haml 4 - which is a bit long in the tooth (Haml::Filter::Coffee class vs. module) #1841

### DIFF
--- a/lib/brakeman/parsers/haml_embedded.rb
+++ b/lib/brakeman/parsers/haml_embedded.rb
@@ -25,20 +25,41 @@ RUBY
   end
 end
 
-# Fake CoffeeScript filter for Haml
-module Haml::Filters::Coffee
-  include Haml::Filters::Base
-  extend Brakeman::FakeHamlFilter
-end
+# Haml::Filters::Coffee used to be defined as a module - now it is a class
+if defined?(Haml::Filters::Coffee) && Haml::Filters::Coffee.class == Class
+  # Fake CoffeeScript filter for Haml
+  class Haml::Filters::Coffee
+    include Haml::Filters::Base
+    extend Brakeman::FakeHamlFilter
+  end
 
-# Fake Markdown filter for Haml
-module Haml::Filters::Markdown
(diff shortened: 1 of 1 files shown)


presidentbeef:
Hmmm... I kind of think if Haml has these filters installed by default now, that the fake filters aren't needed? :thinking: And I _think_ they were added in 6.0 when Hamlit became Haml?

If I recall correctly, the fake filters were just to stub out the embedding of other languages and avoid raising errors. It wasn't to override the filters.

kwerle:
Specifically, we would like to use brakeman-min (6.1.2 or so) with haml (6.3.0) and we can't because brakeman refers to these constants as modules and haml declares them as classes.  I can't think of any solution to this other than the one I've proposed (short of forking brakeman - which seems ridiculous).

Our version of brakeman-min is a little behind, but this is the most recent version of haml.

presidentbeef:
I might have to look again, but I believe I was thinking instead of defining the classes `if defined?(Haml::Filters::Coffee) && Haml::Filters::Coffee.class == Class` to only define the modules if the classes aren't available.

In other words, for newer Haml Brakeman doesn't need to do anything, just skip defining the modules.

kwerle:
> In other words, for newer Haml Brakeman doesn't need to do anything, just skip defining the modules.

I see!  I can take a poke at that.

kwerle:
OK, the good news is that I think that will all work.  The .. bad? new is that brakeman is locked to haml 5.1 so I'm having to do some other gymnastics to work around this issue anyway.  Mostly it means no longer including brakeman in our gemfile (probably the right thing to do) and just run it from a container.

So... victory?  I guess merge this if you think it's the right thing.

And thank you!

presidentbeef:
Yes, looks like there are some breaking changes with Haml 6.x that need to be addressed.

presidentbeef:
Needed to go a slightly different way, but this should be addressed with #1944 finally

PR #1846, opened 2024-05-23
Initial work for adding Rails 8 support

ron-shinall:
This PR adds support for Rails 8.

Fixes https://github.com/presidentbeef/brakeman/issues/1849


ron-shinall:
@presidentbeef I'm not sure when you typically add support for newer Rails versions. Rails 8 alpha officially began 5/13 so I thought I'd take a stab at some initial work on this. I'm not sure of the process you use for adding specific tests - I am happy to  help if you can provide guidance.

presidentbeef:
Hi @ron-shinall, thanks for this.

The reason I hadn't gotten back to you was because I was confused why there would be vulnerabilities in the new application. It looks like these were copied over? Normally I start off with a clean application.

I recommend removing the old vulnerabilities and then I can merge this in. Thanks!

ron-shinall:
@presidentbeef no worries at all. To clarify - are you saying generate a bare skeleton app in `test/apps/rails8`? Also, in `test/tests/rails8.rb` should the `def expected` be removed?

ron-shinall:
@presidentbeef thank you for the input -- please let me know if this is more like what you have in mind.

ron-shinall:
@presidentbeef let me know if there are any more changes needed on this.

presidentbeef:
Thanks @ron-shinall!

PR #1847, opened 2024-06-08
Enable aliases for YAML safe_load of `secrets.yml` file

chaadow:
While checking logs for my brakeman run I've noticed this notice:
`Brakeman::CheckSessionSettings: Unable to parse /Users/c/code/rails_app/config/secrets.yml`

After running the `safe_load` without the `begin..rescue` part, I've had this exception:
```
 Alias parsing was not enabled. To enable it, pass `aliases: true` to `Psych::load` or `Psych::safe_load`. (Psych::AliasesNotEnabled)
```

After enabling the `aliases: true`, the YAML loading worked as expected

### DIFF
--- a/lib/brakeman/checks/check_session_settings.rb
+++ b/lib/brakeman/checks/check_session_settings.rb
@@ -118,7 +118,7 @@ class Brakeman::CheckSessionSettings < Brakeman::BaseCheck
       yaml = secrets_file.read
       require 'yaml'
       begin
-        secrets = YAML.safe_load yaml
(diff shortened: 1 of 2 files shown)


presidentbeef:
Thanks @chaadow - any chance you can update one of the test app's config file so we have a test?

chaadow:
@presidentbeef Yes, let me know if this is what you meant. I picked the most recent secrets.yml ( rails 5.2 )

presidentbeef:
Thanks!

PR #1852, opened 2024-06-19
Add remediation advice for command injection warnings

rangerscience:
Spent far to long pursuing other resolutions (none of which worked, not really) when the answer was `shellescape` all along.


presidentbeef:
Thanks!

But should really be changed over on https://github.com/presidentbeef/brakeman-site if you have the time.

PR #1853, opened 2024-06-20
Fix all warnings in the test suite

casperisfine:
None are really critical to be honest, but it's just good hygienne and makes it easier to spot when a new one is introduced.

### DIFF
--- a/gem_common.rb
+++ b/gem_common.rb
@@ -1,6 +1,7 @@
 module Brakeman
   module GemDependencies
     def self.dev_dependencies spec
+      spec.add_development_dependency "csv"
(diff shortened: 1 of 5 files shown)


presidentbeef:
Thanks!

Haven't seen aliasing a method to itself before...

casperisfine:
Thanks for the merge

> Haven't seen aliasing a method to itself before...

It's an old trick, perhaps you've seen `silence_redefinition_of_method` from Active Support: https://github.com/rails/rails/blob/19eebf6d33dd15a0172e3ed2481bec57a89a2404/activesupport/lib/active_support/core_ext/module/redefine_method.rb#L7

PR #1854, opened 2024-06-25
README.md: `require: false` to bundler instructions

runephilosof-abtion:
No reason to load the code.
Not loading the code is faster.

Brakeman is always executed using `bundle exec brakeman` or using its binstub, which will still work with `require: false`

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Using Bundler:
 
 ```ruby
 group :development do
-  gem 'brakeman'
+  gem 'brakeman', require: false
 end
 ```
 


presidentbeef:
Hi @runephilosof-abtion,

While this is true, it should be noted that I gave up on recommending this a long time ago (#889?) and instead most of Brakeman is dynamically loaded on use. You'll note https://github.com/presidentbeef/brakeman/blob/main/lib/brakeman.rb requires almost no other files.

So the difference between using `require: false` or not is pretty small.

But anyway no harm in adding this for those who pay attention.

PR #1855, opened 2024-06-27
Fix compatibility with `--enable-frozen-string-literal`

casperisfine:
Since Ruby 2.3, Ruby can be started with this options, and it's expected to become the default in Ruby 4.0 (no release date yet).

This commit fixes the callsites that assume string literal are mutable.

Note however that a large part of the test suite is still failing, because of two dependencies:

  - ruby2ruby: https://github.com/seattlerb/ruby2ruby/pull/58
  - erubis: (this one is abandoned so can hardly be fixed).

### DIFF
--- a/lib/brakeman/report/report_markdown.rb
+++ b/lib/brakeman/report/report_markdown.rb
@@ -27,7 +27,7 @@ class Brakeman::Report::Markdown < Brakeman::Report::Table
   end
 
   def generate_report
-    out = "# BRAKEMAN REPORT\n\n" <<
+    out = +"# BRAKEMAN REPORT\n\n" <<
     generate_metadata.to_s << "\n\n" <<
     generate_checks.to_s << "\n\n" <<
     "### SUMMARY\n\n" <<
--- a/lib/brakeman/warning.rb
+++ b/lib/brakeman/warning.rb
@@ -317,7 +317,7 @@ class Brakeman::Warning
 
   def format_ruby code, strip
     formatted = Brakeman::OutputProcessor.new.format(code)
-    formatted.gsub!(/(\t|\r|\n)+/, " ") if strip
+    formatted = formatted.gsub(/(\t|\r|\n)+/, " ") if strip
     formatted
   end
 end
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
       - run: bundle check || bundle install
       - run:
           command: |
-            bundle exec rake
+            bundle exec rake RUBYOPT='--enable-frozen-string-literal --debug-frozen-string-literal'
             ./tmp/cc-test-reporter format-coverage -t simplecov -o tmp/codeclimate.json coverage/.resultset.json
       - store_test_results:
           path: test-results
--- a/gem_common.rb
+++ b/gem_common.rb
@@ -12,7 +12,7 @@ module Brakeman
       spec.add_dependency "parallel", "~>1.20"
       spec.add_dependency "ruby_parser", "~>3.20.2"
       spec.add_dependency "sexp_processor", "~> 4.7"
-      spec.add_dependency "ruby2ruby", "~>2.4.0"
+      spec.add_dependency "ruby2ruby", "~>2.5.1"
       spec.add_dependency "racc"
     end
 
@@ -21,7 +21,7 @@ module Brakeman
       spec.add_dependency "highline", "~>3.0"
       spec.add_dependency "erubis", "~>2.6"
       spec.add_dependency "haml", "~>5.1"
-      spec.add_dependency "slim", ">=1.3.6", "<=4.1"
+      spec.add_dependency "slim", ">=1.3.6", "< 5.3"
       spec.add_dependency "rexml", "~>3.0"
     end
   end
--- a/lib/brakeman/app_tree.rb
+++ b/lib/brakeman/app_tree.rb
@@ -50,7 +50,7 @@ module Brakeman
           "#{Regexp.escape f}\\z"
         end
       end
-      Regexp.new("(?:" << path_regexes.join("|") << ")")
+      Regexp.new("(?:#{path_regexes.join("|")})")
     end
     private_class_method(:regex_for_paths)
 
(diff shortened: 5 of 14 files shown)


lib/brakeman/warning.rb
```diff
@@ -53,6 +53,7 @@ def initialize options = {}
     OPTIONS.each do |key, var|
       self.instance_variable_set(var, options[key])
     end
+    @warning_type = @warning_type.dup
```
presidentbeef, line 56: Most of these look okay, but I'm a bit surprised by this one. I don't expect `warning_type` to get modified. Any clues to where that might be happening?

casperisfine: It fails two tests:

```
  1) Error:
TestReportGeneration#test_tabs_sanity:
FrozenError: can't modify frozen String: "Cross-Site Scripting", created at /Users/byroot/src/github.com/casperisfine/brakeman/lib/brakeman/checks/check_content_tag.rb:199
    lib/brakeman/report/report_tabs.rb:12:in `gsub!'
    lib/brakeman/report/report_tabs.rb:12:in `block (2 levels) in generate_report'
    lib/brakeman/report/report_tabs.rb:10:in `map'
    lib/brakeman/report/report_tabs.rb:10:in `block in generate_report'
    lib/brakeman/report/report_tabs.rb:8:in `map'
    lib/brakeman/report/report_tabs.rb:8:in `generate_report'
    lib/brakeman/report.rb:107:in `generate'
    lib/brakeman/report.rb:58:in `format'
    lib/brakeman/report.rb:63:in `method_missing'
    test/tests/report_generation.rb:67:in `test_tabs_sanity'

  2) Error:
TestTabsOutput#test_reported_warnings:
FrozenError: can't modify frozen String: "Cross-Site Request Forgery", created at /Users/byroot/src/github.com/casperisfine/brakeman/lib/brakeman/checks/check_csrf_token_forgery_cve.rb:19
    lib/brakeman/report/report_tabs.rb:12:in `gsub!'
    lib/brakeman/report/report_tabs.rb:12:in `block (2 levels) in generate_report'
    lib/brakeman/report/report_tabs.rb:10:in `map'
    lib/brakeman/report/report_tabs.rb:10:in `block in generate_report'
    lib/brakeman/report/report_tabs.rb:8:in `map'
    lib/brakeman/report/report_tabs.rb:8:in `generate_report'
    lib/brakeman/report.rb:107:in `generate'
    lib/brakeman/report.rb:58:in `format'
    lib/brakeman/report.rb:63:in `method_missing'
    test/tests/tabs_output.rb:9:in `setup'
```

```ruby
class Brakeman::Report::Tabs < Brakeman::Report::Table
  def generate_report
    [[:generic_warnings, "General"], [:controller_warnings, "Controller"],
      [:model_warnings, "Model"], [:template_warnings, "Template"]].map do |meth, category|

      self.send(meth).map do |w|
        line = w.line || 0
        w.warning_type.gsub!(/[^\w\s]/, ' ')
        "#{(w.file.absolute)}\t#{line}\t#{w.warning_type}\t#{category}\t#{w.format_message}\t#{w.confidence_name}"
      end.join "\n"

    end.join "\n"

  end
end
```

It very explicitly mutate that string in place, hence why I chose to make it mutable. But to be honest it looks like a bug, but I don't know brakeman enough to decide.

If I delete that `gsub!` the test suite still pass.

casperisfine:
About the dependencies

> ruby2ruby: https://github.com/seattlerb/ruby2ruby/pull/58

Looks like @zenspider merged a fix: https://github.com/seattlerb/ruby2ruby/commit/3d5966e30dfe3a04b05e5a6796f6b5d944696001, not sure when we'll get a release thought.


> * erubis: (this one is abandoned so can hardly be fixed).

I could monkey patch the one `erubis` method that isn't compatible, not sure if that'd be OK with you.

zenspider:
I can have a release of r2r out this week.

casperisfine:
> I could monkey patch the one `erubis` method that isn't compatible, not sure if that'd be OK with you.

Ok, so I did it without monkey patch, I just include a module in all of Brakeman's `erubis` based parsers to change the one method.

zenspider:
[r2r is out](https://github.com/seattlerb/ruby2ruby/compare/v2.5.0...v2.5.1)

casperisfine:
thanks @zenspider I updated the dependency, now the entire test suite passes with `--enable-frozen-string-literal`.

lib/brakeman/report/report_markdown.rb
```diff
@@ -27,7 +27,7 @@ def initialize *args
   end
 
   def generate_report
-    out = "# BRAKEMAN REPORT\n\n" <<
+    out = +"# BRAKEMAN REPORT\n\n" <<
```
zenspider, line 30: Personally I really don't like `String#<<`... I wound up changing a lot of my old code from this to `Array#<<` + `#join`. It led to similarly sized diffs, array mutation (which I don't mind) instead of string mutation (which I do), and fits my current style a lot more. 

Not recommending this change here... just something to consider.

zenspider (commented):
Having fixed a lot of these now, I fully approve of this PR. One style note that I eschew `String#<<` which seems in high use here.

lib/brakeman/report/report_markdown.rb
presidentbeef: Indeed... I'm sure there are a lot of improvements that could be made 😄

lib/brakeman/warning.rb
presidentbeef: Let's delete the `gsub!`. Not sure what I thought it was doing... but it doesn't. Also I plan to deprecate+remove that report format.

lib/brakeman/warning.rb
```diff
@@ -175,7 +175,7 @@ def format_user_input strip = true
   def format_with_user_input strip = true, &block
     if self.user_input
       formatted = Brakeman::OutputProcessor.new.format(code, self.user_input, &block)
-      formatted.gsub!(/(\t|\r|\n)+/, " ") if strip
+      formatted = formatted.gsub(/(\t|\r|\n)+/, " ") if strip
```
presidentbeef, line 178: `formatted` is already going to be a fresh new string. Seems wasteful to create another new string right away :thinking:

presidentbeef:
Took some time to get all the testing done, but other than one ruby2ruby output that changed (will follow up on that), looks good! Thanks for your contribution!

PR #1861, opened 2024-07-25
Add `--show-ignored` flag

### LINKED ISSUE #1767: Is there a flag to show all warnings including the ignored ones?
Is there a way to see the ignored warnings without removing their entry from the brakeman.ignore file?

I wanted to see the ignored cases in the cli itself. Is that currently possible?

gazayas:
Closes #1767.

After scaffolding a `Foo` model with a couple of vulnerabilities on a new application, running `brakeman --show-ignored` displays the following:

```
> brakeman --show-ignored
...

== Overview ==

Controllers: 2
Models: 2
Templates: 8
Errors: 0
Security Warnings: 0
Ignored Warnings: 2

== Ignored File Overview ==

Confidence: High
Category: SQL Injection
Check: SQL
Code: Foo.where(("bar = " + params[:bar]))
File: app/controllers/foos_controller.rb
Line: 24

Confidence: High
Category: SQL Injection
Check: SQL
Code: Foo.where(("bar = " + params[:bar]))
File: app/controllers/foos_controller.rb
Line: 40

== Warning Types ==


No warnings found

> echo $?
0
```

This ensures the exit code is unaffected by adding the flag (you can see in the test I added that the exit code returned is `3` which is the default in the command line test).

I wanted to check the output and not just the exit code in the command line test, but I had some trouble finding the right way to do it.

Either way, Looking back at [this comment](https://github.com/presidentbeef/brakeman/issues/1767#issuecomment-1482352588), if we do want to affect the exit code or take another approach, I'd be glad to look over the logic again.

### DIFF
--- a/lib/brakeman/report/report_text.rb
+++ b/lib/brakeman/report/report_text.rb
@@ -103,7 +103,7 @@ class Brakeman::Report::Text < Brakeman::Report::Base
   end
 
   def generate_show_ignored_overview
-    double_space("Ignored File Overview", ignored_warnings.map {|w| output_warning w})
+    double_space("Ignored Warnings", ignored_warnings.map {|w| output_warning w})
   end
 
   def generate_errors
--- a/lib/brakeman/report/report_text.rb
+++ b/lib/brakeman/report/report_text.rb
@@ -9,6 +9,7 @@ class Brakeman::Report::Text < Brakeman::Report::Base
     unless summary_option == :no_summary
       add_chunk generate_header
       add_chunk generate_overview
+      add_chunk generate_show_ignored_overview if tracker.options[:show_ignored] && ignored_warnings.any?
       add_chunk generate_warning_overview
     end
 
@@ -99,6 +100,10 @@ class Brakeman::Report::Text < Brakeman::Report::Base
 
       double_space "Warnings", warnings
     end
+  end
+
+  def generate_show_ignored_overview
+    double_space("Ignored File Overview", ignored_warnings.map {|w| output_warning w})
   end
 
   def generate_errors
--- a/lib/brakeman/report/report_text.rb
+++ b/lib/brakeman/report/report_text.rb
@@ -9,7 +9,6 @@ class Brakeman::Report::Text < Brakeman::Report::Base
     unless summary_option == :no_summary
       add_chunk generate_header
       add_chunk generate_overview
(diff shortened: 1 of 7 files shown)


presidentbeef:
Thanks!

Can you also add a simple test in `test/tests/options.rb`?

gazayas:
Done!

lib/brakeman/report/report_text.rb
```diff
@@ -101,6 +102,10 @@ def generate_warnings
     end
   end
 
+  def generate_show_ignored_overview
+    double_space("Ignored File Overview", ignored_warnings.map {|w| output_warning w})
```
presidentbeef, line 106: This should go at the end of the report, after `generate_warnings`, not as part of the summary. It's not a summary, it's a list of warnings.

presidentbeef: Also title should be "Ignored Warnings" instead of "Ignored File Overview"

gazayas: I see, I went ahead and changed the title here: https://github.com/presidentbeef/brakeman/pull/1861/commits/3203739307155ba34113c958bfee5280412d3389

I moved `generate_show_ignored_overview` to the end below [this line](https://github.com/presidentbeef/brakeman/blob/c83406e7f073ad469cd2f4a44a108cdafd4c6897/lib/brakeman/report/report_text.rb#L23), but I'm getting some test errors so I'll try to push the changes once I figure that part out.

gazayas: Okay, I moved `generate_show_ignored_overview` to the end of the report.

I explicitly return `@output_string` here because adding `generate_show_ignored_overview` at the end was returning `nil` in some cases (When `--show-ignored` isn't used, the `if` statement returns `nil`). All of [these lines](https://github.com/presidentbeef/brakeman/blob/c83406e7f073ad469cd2f4a44a108cdafd4c6897/lib/brakeman/report/report_text.rb#L10-L23)) that use `add_chunk` don't set a second argument so all of the contents are saved to `@output_string`, so I think it should be safe to return the instance variable here explicitly.

presidentbeef: That makes sense. Just coincidence that it's worked all this time since `generate_warnings` always returns something.

PR #1875, opened 2024-10-21
Test with Prism in CI

presidentbeef:


presidentbeef:
Blocked on https://github.com/ruby/prism/issues/3186

PR #1876, opened 2024-10-21
Do not treat private methods as routable

presidentbeef:
When processing controllers, don't try to run before actions and look for templates to render from private methods.

This saves significant time on some applications.


presidentbeef:
Caused more problems than it fixed :shrug:

PR #1891, opened 2024-12-17
Update Prism version to 1.0+

presidentbeef:
Weird the test was updated but not the `--prism` option.


gstokkink:
@presidentbeef can we get a new release with this fix? We'd like to use the `prism` parser as we have a very large code base, and the current parser is too slow 😄 Thanks in advance!

PR #1897, opened 2024-12-29
Use Prism parser by default

presidentbeef:
Fall back to RubyParser if Prism is unavailable or fails

### DIFF
--- a/test/tests/rails8.rb
+++ b/test/tests/rails8.rb
@@ -7,7 +7,7 @@ class Rails8Tests < Minitest::Test
   def report
     @@report ||=
       Date.stub :today, Date.parse("2024-05-13") do
-        BrakemanTester.run_scan "rails8", "Rails 8", run_all_checks: true, use_prism: true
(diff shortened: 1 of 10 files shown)


test/tests/rails8.rb
```diff
@@ -7,7 +7,7 @@ class Rails8Tests < Minitest::Test
   def report
     @@report ||=
       Date.stub :today, Date.parse("2024-05-13") do
-        BrakemanTester.run_scan "rails8", "Rails 8", run_all_checks: true, use_prism: true
+        BrakemanTester.run_scan "rails8", "Rails 8", run_all_checks: true, use_prism: false
```
chaadow, line 10: is it intentional to disable prism here for rails8?

presidentbeef: Yes, it was intentional because Prism is the default now. Switching this to _not_ use prism was an easy way to make sure there's a test for the non-default case.

PR #1899, opened 2024-12-30
Update `eval` check to be a little noisier

presidentbeef:
Previously, the check for dynamic code evaluation (use of `eval`, etc.) only warned if there was definitely user-controllable values being executed.

This PR updates the check to (almost) always warn on use of `eval`, and to warn of use of `class_eval`, `instance_eval`, etc. when there is a dynamic string involved.

Also, the warning message is updated to be a little more specific.


sshaw:
I understand the prior case: 

> only warned if there was definitely **user-controllable values** being executed.

But what's the use case for the expansing? Unless there's a "user-controllable value" this will flag many common use cases for `class_eval` & co. that I would assume are otherwise safe. For example: 

```rb
METHODS = %w[foo bar].freeze
METHODS.each do |name|
  class_eval <<~METHOD, __FILE__, __LINE__ + 1
    def #{name}?
      false
    end
  METHOD
end  
```

Or am I missing something? 

Is `METHODS[n]` "user-controllable" since one can do `METHODS[n].repace "SOMETHING NAUGHTY"`?

presidentbeef:
@sshaw In the case that you shared is the `class_eval` inside of a method definition or at the class level?

Mainly asking because right now if it's in the class definition unfortunately it's a harder fix :disappointed:

sshaw:
They exist inside method definitions and at the class level. 


What's the use case for the expansion? Unless there's a "user-controllable value" this will flag many common use cases for class_eval & co. that I would assume are otherwise safe.

Issue #1738, opened 2022-11-09
safe-methods: pass "pointed" methods?


akimd:
Hi,

I'm using `Sanitize.fragment(str)` to sanitize, but `--safe-methods=Sanitize.fragment` does not work, I have to decay this into `--safe-methods=fragment`.  But I feel less secure this way.
It would be nice to have richer means to specify the safe methods.

Cheers!

presidentbeef:
:thinking: I think this would not be too hard to support.

PR #1916, opened 2025-01-27
Add support for namespaced safe_methods option

evpgh:
Based on the issue or feature request https://github.com/presidentbeef/brakeman/issues/1738 I created a safe method handler to do the parsing of the command line argument `--safe-methods` and comparisons in checks.


evpgh:
Hey @presidentbeef I know this is a low priority but could you take a look?

presidentbeef:
I'm not sure this will work, because uses of `SafeMethodHandler.include?` are only only passed the method name, never the class name. So if `Something.ignore` is the option, it will only ever be compared against `ignore`.

PR #1921, opened 2025-02-24
Add new flag to detect obsolete ignore entries

viralpraxis:
I'd like to propose a new CLI flag `--ensure-no-obsolete-config-entries`. If at least one obsolete (unused) fingerprint is found, brakeman should fail (with exit code 9).

This flag would be mostly useful in e.g. CI environments to ensure code changes are reflected in configuration changes.

### DIFF
--- a/lib/brakeman/commandline.rb
+++ b/lib/brakeman/commandline.rb
@@ -145,6 +145,11 @@ module Brakeman
           quit Brakeman::Errors_Found_Exit_Code
         end
 
+        if tracker.options[:ensure_no_obsolete_ignore_entries] && tracker.unused_fingerprints.any?
+          warn '[Error] Obsolete ignore entries were found, exiting with an error code.'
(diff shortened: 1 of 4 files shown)


viralpraxis:
hey @presidentbeef, could you please take a look?

lib/brakeman/commandline.rb
```diff
@@ -145,6 +145,13 @@ def regular_report options
           quit Brakeman::Errors_Found_Exit_Code
         end
 
+        if tracker.options[:ensure_no_obsolete_ignore_entries] && tracker.unused_fingerprints.any?
+          warn '[Error] The following obsolete ignore entries were found:'
```
presidentbeef, line 149: Obsolete ignore entries are already included in the output, so this shouldn't be necessary.

viralpraxis: Yep, makes sense. Fixed.

viralpraxis:
Sorry for another ping @presidentbeef, can we proceed with this one?

PR #1930, opened 2025-04-01
Allow model paths for classifying files to be configured

Catsuko:
I have some situations where models are not located in standard directories and also use subclasses of `ApplicationRecord` so these files do not get classified as models. Would it be possible to allow the expected model path to be configurable?

This PR shows how I imagine it would work in a backwards compatible way while also allowing the configuration to take advantage of pattern matching.

### Example Usage

```yaml
# brakeman.yml
---
:model_paths:
- app/models/**/*
- lib/packages/*/models/**/*.rb
```

### DIFF
--- a/lib/brakeman.rb
+++ b/lib/brakeman.rb
@@ -52,6 +52,7 @@ module Brakeman
   #  * :interprocedural - limited interprocedural processing of method calls (default: false)
   #  * :message_limit - limit length of messages
   #  * :min_confidence - minimum confidence (0-2, 0 is highest)
+  #  * :model_paths - array of paths for files to be considered as models (default: app/models/**/*)
   #  * :output_files - files for output
   #  * :output_formats - formats for output (:to_s, :to_tabs, :to_csv, :to_html)
   #  * :parallel_checks - run checks in parallel (default: true)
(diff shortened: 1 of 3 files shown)


presidentbeef:
Trying to move away from directory-based file type guessing.

> use subclasses of `ApplicationRecord`

I think this should work if the class hierarchy is available. Do you have an example?

Catsuko:
@presidentbeef 
Thanks for the quick reply, I have something this in `packages/platform/models`:

```ruby
class DatabaseABase < ActiveRecord::Base
end

class DatabaseBBase < ActiveRecord::Base
end

class Message < DatabaseABase
end

class User < DatabaseBBase
end
```

Which define connection details for some different databases (following the [rails guide here](https://guides.rubyonrails.org/active_record_multiple_databases.html)). When running brakeman with some logging, I can see that `DatabaseABase` and `DatabaseBBase` are classified as models but `User` and `Message` are not.

When I briefly looked at how the class hierarchy is checked [here](https://github.com/presidentbeef/brakeman/blob/main/lib/brakeman/processors/lib/file_type_detector.rb#L31) it seemed to only check the immediate parent rather than all ancestors.

Catsuko:
@presidentbeef 
Alternatively what if I could configure the base classes to look for? I can't see a clear way to check class hierarchy when classifying files 🤔

presidentbeef:
Perhaps using [`ancestor?`](https://github.com/presidentbeef/brakeman/blob/753e54d253bab57a5d87bbfb31ee2ad97593fc86/lib/brakeman/tracker/collection.rb#L25)

Might have to pull that method out to re-use.

PR #1933, opened 2025-04-15
Consider Tempfile.create.path as safe input

aliismayilov:
This PR adds extends the base check to consider `Tempfile.create.path` to be a safe input. This was already the case for `Tempfile.new`. According to ruby docs `Tempfile.create` is the recommended usage: https://docs.ruby-lang.org/en/3.4/Tempfile.html

### DIFF
--- a/lib/brakeman/checks/base_check.rb
+++ b/lib/brakeman/checks/base_check.rb
@@ -151,10 +151,13 @@ class Brakeman::BaseCheck < Brakeman::SexpProcessor
     method[-1] == "?"
   end
 
-  TEMP_FILE_PATH = s(:call, s(:call, s(:const, :Tempfile), :new), :path).freeze
+  TEMP_FILE_PATH = [
+    s(:call, s(:call, s(:const, :Tempfile), :new), :path).freeze,
+    s(:call, s(:call, s(:const, :Tempfile), :create), :path).freeze
+  ].freeze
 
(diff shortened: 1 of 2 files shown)


aliismayilov:
I couldn't figure out how to reliably match `Tempfile.create` yielded argument, so that I could make the following to be ignored as well:
```ruby
Tempfile.create do |tempfile|
  system("something -out #{tempfile.path}")
end
```

Any help would be appreciated!

presidentbeef:
For `Tempfile.create` with a block, there is very similar code for `Tempfile.open` [here](https://github.com/presidentbeef/brakeman/blob/3b99363c80bc651e38c3beb511018b17f80dc465/lib/brakeman/processors/alias_processor.rb#L465-L467) that you can adjust/adapt.

aliismayilov:
> For `Tempfile.create` with a block, there is very similar code for `Tempfile.open` [here](https://github.com/presidentbeef/brakeman/blob/3b99363c80bc651e38c3beb511018b17f80dc465/lib/brakeman/processors/alias_processor.rb#L465-L467) that you can adjust/adapt.

Thanks for the pointer! I updated the PR to cover `Tempfile.create` block as well

aliismayilov:
The tests pass, but the tests CI step fails due to codeclimate results upload 🤷

PR #1934, opened 2025-04-17
Clean-up and update JUnit report to make it compatible with CircleCI

pbernery:
The report was adding custom XML tags and attributes to add more info in the output, however this is not supported by some Continuous Integration services like CircleCI.

This PR removes those custom attributes, but also changes the XML format slightly: `file` and `line` attributes are used to report code which breaks, which should create more elegant and convenient reports on CI tools.

This PR also removes some XML tags that was added to the report without content.

It has been tested on CircleCI which currently shows the report like this:


presidentbeef:
Great, thanks!

PR #1936, opened 2025-04-23
Check for unsafe SQL when two arguments are passed to AR methods

patbl:
I think this conditional is meant to

1. Always check the first and last arguments.
2. Avoid checking the argument twice when there's only one argument.

The structure of `call` appears to be:

* 0: the symbol `:call`
* 1: the receiver of the method call
* 2: the method being called
* 3: the first argument
* 4: the second argument
* 5: the third argument (and so on)

So `call.length > 5` will be true only if there are at least 3 arguments.

Before this change, the behavior was

```ruby
User.count("#{params[:input]}") # triggers warning
User.count("#{params[:input]}", :foo) # doesn't trigger warning
User.count("#{params[:input]}", :foo, :bar) # triggers warning
```

After this change, the two-argument version will trigger the warning.

Another option would be to use `if call.arglist.length > 1`. This is more explicit, but it involves an additional method call, and I don't know what the performance implications of that are.

### DIFF
--- a/lib/brakeman/checks/check_sql.rb
+++ b/lib/brakeman/checks/check_sql.rb
@@ -192,7 +192,7 @@ class Brakeman::CheckSQL < Brakeman::BaseCheck
                       when :last, :first, :all
                         check_find_arguments call.first_arg
                       when :average, :count, :maximum, :minimum, :sum
-                        if call.length >= 5 # more than one argument was passed
+                        if call.arglist.length > 1
                           unsafe_sql?(call.first_arg) or check_find_arguments(call.last_arg)
                         else
                           check_find_arguments call.last_arg
--- a/lib/brakeman/checks/check_sql.rb
+++ b/lib/brakeman/checks/check_sql.rb
@@ -192,7 +192,7 @@ class Brakeman::CheckSQL < Brakeman::BaseCheck
                       when :last, :first, :all
                         check_find_arguments call.first_arg
                       when :average, :count, :maximum, :minimum, :sum
-                        if call.length > 5
+                        if call.length >= 5 # more than one argument was passed
                           unsafe_sql?(call.first_arg) or check_find_arguments(call.last_arg)
                         else
                           check_find_arguments call.last_arg


presidentbeef:
I'm going to have to figure out what the original intent was here... it's been a while since those methods even accepted more than one argument.

(But I think you are right)

presidentbeef:
@patbl Can you update to use the more explicit `call.arglist.length > 1`?

patbl:
@presidentbeef I made that change.

soulcutter:
This change seems to have triggered a LOT of warnings across many of our applications that use a `statsd`-like interface for calling `$stats.count("metric.name.#{custom_thing}", 1)` - something we use WAY more than ActiveRecord count-with-2-args.

https://www.rubydoc.info/gems/dogstatsd-ruby/3.0.0/Datadog/Statsd#count-instance_method

Is there a way this check might be better-targeted to avoid this deluge of false positives?

presidentbeef:
@soulcutter can you provide an example of the exact warning from Brakeman? Does it say `$stats` or does it know what `$status` is?

soulcutter:
```
Confidence: Weak
Category: SQL Injection
Check: SQL
Code: $stats.count("prefix.#{variable}", :tags => ({ :result => "error_raised" }))
File: app/models/<model>.rb
Line: 61
```

The `$stats` global is initialized in dependency, not defined anywhere in the application code (it's not defined in `app/models/<model>.rb` for example). I do not know how to answer the question of if it knows what `$stats` is

presidentbeef:
Maybe Brakeman can just ignore calls that look like ActiveRecord methods but are on globals :thinking:

patbl:
> Maybe Brakeman can just ignore calls that look like ActiveRecord methods but are on globals 🤔

A global variable such as `$statsd` is one way of referencing this sort of object, but applications could instead use a module, class, method, constant, or local variable. (`StatsD`, `STATSD`, `statsd`...).

It would be nice if there were a way to tell Brakeman "this [`$statsd`, `STATSD`, etc.] isn't an Active Record class." But maybe that's too much scope creep (I'm assuming something like this doesn't already exist).

presidentbeef:
Writing down for myself... for common methods like `count` this check can go back to only warning if the target is known to be an ActiveRecord model.

PR #1938, opened 2025-05-12
Only load from `cgi` what is required

Earlopain:
In Ruby 3.5 most of the `cgi` gem will be removed. Only the various escape/unescape methods will be retained by default.

https://bugs.ruby-lang.org/issues/21258


presidentbeef:
Thanks!

Issue #1945, opened 2025-07-08
Brakeman enum support broken?


cubasepp:
### Background

Brakeman version: 7.0.2
Rails version: 7.2.2.1
Ruby version: 3.4.4

Link to Rails application code: ?

### Issue

When I use an enum like:
```
self.where("thing IN #{Model.statuses[:value]}")
```
I get a Possible SQL injection Waning.

I believe the `add_enum_method` is not up to date? [code](https://github.com/presidentbeef/brakeman/blob/3b99363c80bc651e38c3beb511018b17f80dc465/lib/brakeman/processors/model_processor.rb#L93)

If step into this function for some reason the first two checks are false:

```
return unless hash? arg
return unless symbol? arg[1]
```

If it's really a bug I can try to fix it, can you confirm it's broken?

#### Other Error

Run Brakeman with `--debug` to see the full stack trace.

Stack trace:

```
Confidence: High
Category: SQL Injection
Check: SQL
Message: Possible SQL injection
Code: self.where("thing IN #{Model.statuses[:value]}")
File: app/models/model.rb
Line: 
```

presidentbeef:
Yes... looks like the enum API changed in Rails 7.1.

cubasepp:
I tried to provide a fix.
But I am not sure if I understand the test setup correct. 
Maybe you can give me some advice how to fix the old test! Thx.

PR #1946, opened 2025-07-09
Fix enum support

cubasepp:
[Issue 1945](https://github.com/presidentbeef/brakeman/issues/1945)

In newer rails version brakeman does not recognize enums as safe.

### DIFF
--- a/lib/brakeman/processors/model_processor.rb
+++ b/lib/brakeman/processors/model_processor.rb
@@ -90,13 +90,23 @@ class Brakeman::ModelProcessor < Brakeman::BaseProcessor
     end
   end
 
-  def add_enum_method call
-    arg = call.first_arg
-    return unless hash? arg
-    return unless symbol? arg[1]
+  def add_enum_method(call)
+    if call.second_arg.nil?
+      arg = call.first_arg
+      return unless hash? arg
+      return unless symbol? arg[1]
 
-    enum_name = arg[1].value # first key
-    enums = arg[2] # first value
+      enum_name = arg[1].value # first key
+      enums = arg[2] # first value
+    else
+      first_arg = call.first_arg
+      second_arg = call.second_arg
+      return unless symbol? first_arg
+      return unless (hash?(second_arg) || array?(second_arg))
+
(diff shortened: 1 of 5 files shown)


lib/brakeman/processors/model_processor.rb
```diff
@@ -90,13 +90,25 @@ def process_call exp
     end
   end
 
-  def add_enum_method call
-    arg = call.first_arg
-    return unless hash? arg
-    return unless symbol? arg[1]
+  def add_enum_method(call)
+    if rails_version >= '7.1'
```
cubasepp, line 94: Not sure how to deal with different versions?

presidentbeef:
Brakeman still needs to support the older API for enums.

I'm not sure why you moved the Gemfile for the Rails 7 test app, there shouldn't be any need to touch it.

cubasepp:
> Brakeman still needs to support the older API for enums.
> 
> I'm not sure why you moved the Gemfile for the Rails 7 test app, there shouldn't be any need to touch it.

I wasn’t really sure how the whole thing with the tests works. Then I had a look at it and thought Rails 7 is different compared to the other apps.
Anyway, I’ve now just added the test cases from Rails 6.

I think the new code isn’t being tested. When I set a debugger in the Rails 8 app, it still shows the old behavior. I’m not sure how this is supposed to work. In my app, it works fine.

Could it be that the [to_test.rb](https://github.com/presidentbeef/brakeman/blob/main/test/to_test.rb) is outdated as well?
I was wondering — isn’t it possible to generate the test files automatically?

presidentbeef:
No need to add duplicate tests to different test apps.

You've only added new instances of the old-style API, so that's why the new code is not being exercised.

`to_test.rb` simply generates tests based on the warnings emitted. So to test a fixed false positive, what I do is add the false positive to the test app and run `to_test.rb` without the fix. That generates an `assert_warning ...`. That can then be changed to `assert_no_warning` - which will fail at first. Then add in your fix and it should pass.

cubasepp:
Thanks for walking me through that.

PR #1955, opened 2025-07-28
[#1942]Add support for Rails configuration modules

### LINKED ISSUE #1942: Allow Brakeman to determine the Rails configuration by querying `Rails.application.config`

We have a bunch of Rails applications in a single repository (i.e., a monorepo). These Rails applications share a lot of the same configuration. We want to manage this shared configuration in a single place, thus we include shared modules in various configuration files (for instance, `application.rb`). Because Brakeman currently parses the configuration files directly, it misses the configuration that's set by the included modules. For instance, if we were to set `config.action_controller.default_protect_from_forgery = true` in a shared configuration module rather than setting it directly in `application.rb` (or `environment.rb`), Brakeman would not detect this, and will start complaining about controllers missing forgery protection. If Brakeman were to query `Rails.application.config.action_controller.default_protect_from_forgery` instead, however, it would see that it's set.

Brakeman to determine the Rails configuration by querying `Rails.application.config` instead of parsing the configuration files.

**Describe alternatives you've considered**

Including the shared configuration modules in the files that Brakeman scans. This requires hacking the Brakeman internals, however. Perhaps an option to define additional places where configuration is defined?

**Additional context**

gstokkink-bex:
Just a start to see if this approach is feasible.

Closes #1942


gstokkink-bex:
@presidentbeef what do you think of this approach? It works great for our use case. Haven't added any tests yet because I'd first like to know if you think this is a good approach or not 😋

presidentbeef:
Seems reasonable!

presidentbeef:
Still needs tests.

gstokkink-bex:
Yes, hopefully will have some time soon!

gstokkink-bex:
@presidentbeef I've added tests based on the existing Rails 8 tests 😄 Part of the config is now moved to a separately included config file. Everything keeps working, as expected. Is this sufficient?

gstokkink-bex:
Checks failing are not relevant here, I think 😄

gstokkink-bex:
@presidentbeef just wondering if you have seen this? 😄

presidentbeef:
Oops, this is adding too much.

Instead of adding a whole new app, I'd say add just change the existing Rails 8 app to use the new config file/path. I think you may need to directly test that the configuration in the file was actually picked up as well. E.g., `tracker.config.rails.some.setting == ...`

PR #1963, opened 2025-09-03
Fix SQL injection check for calculate method

rsharma-figma:
## Summary

The calculate method was only checking the third argument (options hash) but not the second argument (column_name) which can also contain user input. This fix ensures both the column_name and options are properly checked for SQL injection vulnerabilities, similar to other aggregate methods.

As documented on https://rails-sqli.org/, ActiveRecord calculation methods like `calculate` are vulnerable to SQL injection when user input is directly passed into column name arguments:

```ruby
# Vulnerable example:
params[:column] = "age) FROM users WHERE name = 'Bob';" 
Order.calculate(:sum, params[:column])
```

This fix addresses the vulnerability by checking both the second argument (column_name) and third argument (options) for SQL injection patterns.

## Changes

- Updated SQL injection check for `calculate` method in `lib/brakeman/checks/check_sql.rb`
- Added proper argument length checks similar to other aggregate methods
- Added test case for SQL injection in calculate method column_name parameter
- Updated expected warning count and line numbers in test files

## Test Plan

- All existing tests pass
- New test case verifies SQL injection detection in calculate method column_name
- The calculate method test cases now properly detect SQL injection in both column_name and options arguments

Fixes SQL injection detection gap in calculate method as referenced in https://rails-sqli.org/.

### DIFF
--- a/test/apps/rails3.1/app/models/product.rb
+++ b/test/apps/rails3.1/app/models/product.rb
@@ -137,6 +137,7 @@ class Product < ActiveRecord::Base
   def test_calculations
     #Should warn
     Product.calculate(:count, :all, :conditions => "blah = '#{params[:blah]}'")
+    Product.calculate(:sum, params[:column_name])  # Should warn - column name from params
     Product.minimum(:price, :conditions => "blah = #{params[:blach]}")
     Product.maximum(:price, :group => params[:columns])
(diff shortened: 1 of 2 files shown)


presidentbeef:
Thanks!

Pretty embarrassing, since I made that website...

Issue #1952, opened 2025-07-18
Should dom_id be a default XSS safe method?


jaredmoody:
### Background

Brakeman version: 7.1.0
Rails version: 7.2.2.1
Ruby version: 3.3.8

Just upgraded to brakeman 7.1 (thank you!) and getting a lot of new Cross-Site Scripting/Unescaped model attributes for code like: `%li{ id: dom_id(Item.find(1)) }` Wondering if `dom_id` should be a safe method since it uses the id of the model which (usually?) isn't user-settable.

presidentbeef:
Yes, it should be.

jaredmoody:
I can submit a PR to add that method, but thinking about this some more...

- Looking at the [implementation](https://github.com/rails/rails/blob/3235827585d87661942c91bc81f64f56d710f0b2/actionview/lib/action_view/record_identifier.rb#L93), if an unsafe prefix were passed, I think `dom_id` could be unsafe? E.g. `dom_id(User.first, User.first.my_user_settable_attribute)`. It looks like it's just interpolating the prefix arg into the string.
- `dom_class` came up a lot for me too, and `dom_id`, it calls `dom_class` internally - so `dom_class` would have to be safe if `dom_id` is.

mjankowski:
On this version bump PR - https://github.com/mastodon/mastodon/pull/35434 - we also see a similar seeming failure

Example - https://github.com/mastodon/mastodon/actions/runs/16744634916/job/47400165887?pr=35434#step:6:2246

We have various places in the admin area views all using `dom_id` in what feels like a similar way to each other -- [invite](https://github.com/mastodon/mastodon/blob/081d38679f965ac20ff1a6c72e2fc1108c34ac30/app/views/invites/_invite.html.haml#L1) / [admin report_note](https://github.com/mastodon/mastodon/blob/081d38679f965ac20ff1a6c72e2fc1108c34ac30/app/views/admin/report_notes/_report_note.html.haml#L1) / [admin announcement](https://github.com/mastodon/mastodon/blob/081d38679f965ac20ff1a6c72e2fc1108c34ac30/app/views/admin/announcements/_announcement.html.haml#L1) / [admin invite](https://github.com/mastodon/mastodon/blob/081d38679f965ac20ff1a6c72e2fc1108c34ac30/app/views/admin/invites/_invite.html.haml#L1) -- but only some of them (admin announcement, admin report note)  are flagged by the scan.

[...]

PR #1966, opened 2025-10-15
Add `dom_id` and `dom_class` to ignore methods for `CrossSiteScripting`

mjankowski:
Described/requested in https://github.com/presidentbeef/brakeman/issues/1952#issuecomment-3092654010

Because I'm not 100% sure how to reliably trigger these warnings (see note in linked issue), I'm not sure where to add coverage for this. Specifically in this case, the `_user` partial in the test suite DOES call `dom_id`, but does not trigger a warning ... I'm assuming because whatever combination of params/setup from the controller does not cause it to happen. Would love guidance on that if it's needed in this case.

### DIFF
--- a/lib/brakeman/checks/check_cross_site_scripting.rb
+++ b/lib/brakeman/checks/check_cross_site_scripting.rb
@@ -289,13 +289,13 @@ class Brakeman::CheckCrossSiteScripting < Brakeman::BaseCheck
   end
 
   def setup
-    @ignore_methods = Set[:==, :!=, :button_to, :check_box, :content_tag, :escapeHTML, :escape_once,
-                           :field_field, :fields_for, :form_for, :h, :hidden_field,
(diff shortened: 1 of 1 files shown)


presidentbeef:
Thanks for the continued investigation, @mjankowski!

However, I believe the root issue is that `::Haml::AttributeBuilder.build_id` is part of Haml internals that just shouldn't be surfaced at all. And it turns out the pre-Haml 6.x handling in Brakeman already accounted for this.

Can you see if https://github.com/presidentbeef/brakeman/pull/1967 addresses your issue? (I didn't dig up `dom_class` examples, so I'm just assuming they were similar).

mjankowski:
Thanks for quick response.

Yes, when I run https://github.com/presidentbeef/brakeman/pull/1967 against local repo, that also resolves the issue.

Will close this in favor of that one.

PR #1976, opened 2025-11-09
Fix singleton method prefixes

viralpraxis:
In case of something like

```ruby
class A
  class << self
    def foo = ??
  end
end
```

the prefixed name of the `foo` method is `s(:self).foo`, not `self.foo`.

Looks like ideally it should be `A.foo` (?), but this patch does not address that.

Here's how it looks in the report,

before:

```
"location": {
  "type": "method",
  "class": "SomeModule::SomeClass",
  "method": "s(:self).some_singleton_method"
}
```

after:

```
"location": {
  "type": "method",
  "class": "SomeModule::SomeClass",
  "method": "self.some_singleton_method"
}
```

### DIFF
--- a/lib/brakeman/tracker/collection.rb
+++ b/lib/brakeman/tracker/collection.rb
@@ -55,11 +55,21 @@ module Brakeman
       if src.node_type == :defs
         @class_methods[name] = meth_info
 
-        # TODO fix this weirdness
-        name = :"#{src[1]}.#{name}"
+        name = :"#{method_definition_receiver(src[1])}.#{name}"
       end
 
       @methods[visibility][name] = meth_info
+    end
+
+    def method_definition_receiver(receiver)
+      return receiver if receiver.is_a?(Symbol)
+
+      case receiver.sexp_type
+      when :self
+        "self"
+      else
+        receiver[1].to_s
+      end
     end
 
     def each_method
--- a/test/apps/rails8/lib/evals.rb
+++ b/test/apps/rails8/lib/evals.rb
@@ -26,4 +26,23 @@ class Evals
       METHODS
     end
   end
+
+  class << self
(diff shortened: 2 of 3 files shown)


lib/brakeman/tracker/collection.rb
```diff
@@ -56,7 +56,9 @@ def add_method visibility, name, src, file_name
         @class_methods[name] = meth_info
 
         # TODO fix this weirdness
-        name = :"#{src[1]}.#{name}"
+        # See https://github.com/presidentbeef/brakeman/pull/881 for context
+        prefix = sexp?(src[1]) ? src[1][0] : src[1]
+        name = :"#{prefix}.#{name}"
```
presidentbeef: I think this won't work if `src[1]` is not `self`.

E.g.

```ruby
def A.foo
  # ...
end
```

`src[1].value` might work better (I haven't tested thoroughly).

viralpraxis:
Sorry for ping @presidentbeef, any chances to make some progress here?

lib/brakeman/tracker/collection.rb
```diff
@@ -56,12 +56,24 @@ def add_method visibility, name, src, file_name
         @class_methods[name] = meth_info
 
         # TODO fix this weirdness
```
presidentbeef: I think this fixes the weirdness? So you can delete this comment.

presidentbeef:
I think is good to go (with just that one cleanup). It will mean some warning fingerprints change but I'm debating an 8.0 release anyway.

lib/brakeman/tracker/collection.rb
viralpraxis: Done!

presidentbeef:
Thanks!

Issue #1632, opened 2021-08-12
False positive: Format Validation for ruby built-in constant regular expression


nlamantia:
### Background

Brakeman version: 5.1.1
Rails version: 6.0.3.7
Ruby version: 2.7.4

Link to Rails application code: ?
#### False Positive

```
Confidence: High
--
  | [2021-08-11T21:21:06Z] Category: Format Validation
  | [2021-08-11T21:21:06Z] Check: ValidationRegex
  | [2021-08-11T21:21:06Z] Message: Insufficient validation for `initiated_by` using `/[a-zA-Z0-9.!\#$%&'*+\/=?^_`{\|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)/`. Use `\A` and `\z` as anchors
```

Relevant code:

```ruby
validates :initiated_by, :format => { :with => URI::MailTo::EMAIL_REGEXP }
```

_Why might this be a false positive?_

`URI::MailTo::EMAIL_REGEXP` evaluates to ```/\A[a-zA-Z0-9.!\#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*\z/```. So the flagged code does use a regex that is properly anchored, but it keeps popping up as a warning. If I create my own constant for the exact same regex, as below, it is not flagged as a security warning:

```ruby
EMAIL_REGEX = /\A[a-zA-Z0-9.!\#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*\z/
...
validates :intiated_by, :format => { :with => EMAIL_REGEX }
```

presidentbeef:
Hi @naokikimura.

The regex and code you are referencing do not match the warning.

Note the warning is about the `initiated_by` attribute and the regex does not have `\A` or `\z` anchors.

[...]

PR #1981, opened 2025-12-04
Fix qualified constant lookup to respect module/class context

flavorjones:
When looking up qualified constants like `B::MY_CONST`, Brakeman would incorrectly match constants defined in other modules (like `A::MY_CONST`) because unqualified constants were stored without their module/class context.

When storing a simple constant defined inside a module/class, prepend the context to build the fully qualified name. `MY_CONST` defined in module `A` is now stored as the equivalent of `[:A, :MY_CONST]` rather than `[:MY_CONST]`, enabling correct lookups.

This fixes issues like the one described at #1632

### DIFF
--- a/lib/brakeman/tracker/constants.rb
+++ b/lib/brakeman/tracker/constants.rb
@@ -29,7 +29,7 @@ module Brakeman
 
     def set_name name, context
       @name = name
-      @name_array = Constants.constant_as_array(name)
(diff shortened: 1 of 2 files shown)


presidentbeef:
Thank you so much!

flavorjones:
@presidentbeef Thanks for maintaining this gem! ❤️

PR #1987, opened 2025-12-25
Update ruby_parser to handle ruby 4.0.0+

chaadow:
https://rubygems.org/gems/ruby_parser


presidentbeef:
Annoyingly, I tested with 4.0.0-preview3... which is technically "< 4" :roll_eyes:

chaadow:
Yeah same I've tested with `4.0.0-preview3` and thought everything was ready. 

Had the same issue with multiple `faraday` gems as well as this one ( `ruby_parser` )

Thanks for merging quickly!

Issue #2003, opened 2026-02-02
Wrong number of arguments to `logger.cleanup` when filtering warnings in interactive mode


imran-iq:
Called here: https://github.com/presidentbeef/brakeman/blob/406e8f1013d7bff6221184696cc8dc38f7805c93/lib/brakeman.rb#L617

Defined here: https://github.com/presidentbeef/brakeman/blob/406e8f1013d7bff6221184696cc8dc38f7805c93/lib/brakeman/logger.rb#L60

Causes the following error:

when invoked `bundle exec brakeman  -I --only-files not_important.rb`

```
Brakeman v8.0.1
Loading scanner...
Scanning /Volumes/code/carrot/shoppers/shoppers
Processing gems...
Processing configuration...
Finding files...
Parsing files...
Detecting file types...
Processing initializers...
Processing libraries...
Processing routes...
Processing templates...
Processing data flow...
Processing models...
Processing controllers...
Processing data flow...
Indexing method calls...
Running 86 checks in parallel...
Filtering warnings......
bundler: failed to load command: brakeman (/Users/imran/.rbenv/versions/3.3.0/bin/brakeman)
/Users/imran/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/brakeman-8.0.1/lib/brakeman/logger.rb:60:in `cleanup': wrong number of arguments (given 0, expected 1) (ArgumentError)
        from /Users/imran/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/brakeman-8.0.1/lib/brakeman.rb:617:in `block in filter_warnings'
        from /Users/imran/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/brakeman-8.0.1/lib/brakeman/logger.rb:111:in `context'
        from /Users/imran/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/brakeman-8.0.1/lib/brakeman.rb:655:in `process_step'
        from /Users/imran/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/brakeman-8.0.1/lib/brakeman.rb:614:in `filter_warnings'
        from /Users/imran/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/brakeman-8.0.1/lib/brakeman.rb:453:in `scan'
[...]

PR #2004, opened 2026-02-02
Fix argument error to logger.cleanup

imran-iq:
I chose to update the definition to match the Base and Console logger instead of updating the call site.

Fixes: https://github.com/presidentbeef/brakeman/issues/2003


presidentbeef:
Thanks!

PR #2008, opened 2026-02-05
Add age option for --ensure-latest

### LINKED ISSUE #1989: --ensure-latest may not be sound policy anymore
By default, rails sets up brakeman to run in CI with "--ensure-latest".  While that seemed like sound policy before, in the light of some of the supply chain attacks, it's now a lot more reasonable to wait for a week or so after new dependency versions are pushed before updating.   Implementing such a check for brakeman in CI for brakeman is doable, but obnoxious.  Would it be feasible to add a flag that checked if the version of brakeman was more than N days out-of-date before failing?

presidentbeef:
Fixes #1989

### DIFF
--- a/lib/brakeman/options.rb
+++ b/lib/brakeman/options.rb
@@ -63,7 +63,11 @@ module Brakeman::Options
           options[:exit_on_error] = exit_on_error
         end
 
-        opts.on "--ensure-latest [DAYS]", Integer, "Fail when Brakeman is outdated. Optionally set minimum age in days.", (1..15) do |days|
+        opts.on "--ensure-latest [DAYS]", Integer, "Fail when Brakeman is outdated. Optionally set minimum age in days." do |days|
+          if days and not (1..15).include? days
+            raise OptionParser::InvalidArgument
+          end
+
           options[:ensure_latest] = days || true
         end
 


lib/brakeman.rb
```diff
@@ -416,11 +416,25 @@ def self.dump_config options
...
+  def self.ensure_latest(days_old: 0)
     current = Brakeman::Version
-    latest = Gem.latest_version_for('brakeman').to_s
-    if current != latest
-      "Brakeman #{current} is not the latest version #{latest}"
+    latest = Gem.latest_spec_for('brakeman')
+    release_date = latest.date.to_date
+    latest_version = latest.version.to_s
+
+    if (Date.today - latest.date.to_date) >= days_old
```
eliotsykes: What is the desirable behavior for `--ensure-latest` when there are multiple newer versions available, but only some of them are too new?

For example:

1. `days_old` is set to `2`.
2. The locally installed version of brakeman is `99.0.0` (`current`).
3. There are 4 newer versions of brakeman:
  a. `99.0.1` released 4 days ago (is old enough)
  b. `99.0.2` released 3 days ago (also old enough)
  c. `99.0.3` released yesterday (too new)
  d. `99.0.4` released today (also too new)

PR #2010, opened 2026-02-15
Fix `TemplateAliasProcessor#template_name` arity

viralpraxis:
```shell
<snip>/gems/brakeman-8.0.2/lib/brakeman/processors/lib/render_helper.rb:26 - wrong number of arguments (given 0, expected 1)
```

### DIFF
--- a/lib/brakeman/processors/lib/render_helper.rb
+++ b/lib/brakeman/processors/lib/render_helper.rb
@@ -19,7 +19,9 @@ module Brakeman::RenderHelper
       end
     when :default
       begin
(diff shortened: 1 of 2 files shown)


presidentbeef:
I don't think this is the right fix. I think the call in `render_helper` is... wrong? Or something else is going on.

viralpraxis:
> I don't think this is the right fix. I think the call in `render_helper` is... wrong? Or something else is going on.

Yep, the call was wrong. We should always pass the action name as an argument. Fixed

viralpraxis:
@presidentbeef can we proceed with this one?

PR #2019, opened 2026-04-19
Resolve index_by(&:itself) on literal arrays to hash literals

### LINKED ISSUE #2018: Hash lookup on frozen constant should untaint the result
When a tainted value (e.g. `params[:id]`) is used as a key to look up a value from a frozen constant hash, the result is currently treated as tainted. However, the return value can only be one of the hash's constant values (or nil) — user input selects which value is returned but doesn't flow into the result itself.

Example:

```ruby
class MailtemplatesController < ApplicationController
  TEMPLATE_IDS = %w[
    layout message_available password_reset
  ].index_by(&:itself).freeze

  def set_template_id
    @template_id = TEMPLATE_IDS[params[:id]]
  end

  def edit
    # Brakeman flags this as "Parameter value used in file name"
    # but @template_id can only be "layout", "message_available",
    # "password_reset", or nil — never arbitrary user input.
    File.read(File.join(@template_path, "#{@template_id}.html"))
  end
end
```

Brakeman reports this as a File Access warning because it traces taint from `params[:id]` through the hash lookup and instance variable assignment. But `TEMPLATE_IDS[anything]` can only return a value that existed in the hash at freeze time — the tainted key cannot influence what string is returned, only which one.

Suggested behavior:

For `Hash#[]` and `Hash#fetch` where the receiver is a frozen constant defined at the class/module level with all-literal values, the return value should not carry taint from the key.

Edge cases to consider:

`HASH.fetch(tainted)` — safe (raises KeyError on miss)
`HASH.fetch(tainted) { block }` — block return needs its own taint analysis
`HASH.fetch(tainted, default)` — only safe if default is also untainted
Non-frozen or non-constant hashes should continue to propagate taint
Workaround:

Currently using `brakeman.ignore` for these false positives. The pattern of "validate input against a constant allowlist, then use the looked-up value" is common enough that supporting it would reduce ignore-file noise.

I can possibly submit a PR for this change but wanted to open this first to make sure it's agreed on as a good idea before writing anything. Thank you.

allard:
Fixes #2018

## Problem

When a constant is defined as `%w[a b c].index_by(&:itself).freeze`, Brakeman doesn't know what `index_by` does, so the constant's value remains an unresolved `:call` sexp. This means lookups like `LOOKUP[params[:id]]` propagate taint from `params[:id]` through the hash access, even though the return value can only be one of the hash's constant values.

## Solution

Teach Brakeman to convert `array.index_by(&:itself)` into a hash literal (`{"a" => "a", "b" => "b", ...}`) when the array contains all literal values. This is done in two places:

1. **`AliasProcessor#process_call_target`** — handles inline usage within method bodies
2. **`Constants#add`** — handles class-level constant definitions (after `.freeze` is already stripped)

Once the constant resolves to a hash literal, Brakeman's existing logic correctly untaints hash lookups with tainted keys, eliminating false positive File Access, Path Traversal, Cross-Site Scripting, and Dynamic Render Path warnings.

## Scope

Only converts `index_by(&:itself)` — not arbitrary blocks like `index_by(&:to_s)` or `index_by { |x| x.name }`. The conversion requires all array elements to be literals.

## Tests

- 3 alias processor tests: `index_by(&:itself)`, with `.freeze`, and non-`:itself` block pass (no conversion)
- 2 constants tests: direct and with `.freeze` wrapping

## Real-world validation

Tested against a Rails app that defines:
```ruby
TEMPLATE_IDS = %w[layout message_available ...].index_by(&:itself).freeze
```
and uses `TEMPLATE_IDS[params[:id]]` in file paths. Before: 14 false positive warnings. After: 0.

### DIFF
--- a/lib/brakeman/tracker/constants.rb
+++ b/lib/brakeman/tracker/constants.rb
@@ -105,6 +105,15 @@ module Brakeman
         value = value.target
       end
 
+      # Convert %w[...].index_by(&:itself) to a hash literal
+      if call? value and value.method == :index_by and
+          array? value.target and all_literals? value.target and
+          node_type? value.first_arg, :block_pass and
+          value.first_arg[1] == s(:lit, :itself)
+        hash_body = value.target.each_sexp.flat_map { |e| [e, e.deep_clone] }
+        value = s(:hash, *hash_body).line(value.line)
+      end
+
       base_name = Constants.get_constant_base_name(name)
       @constants[base_name] ||= []
       @constants[base_name] << Constant.new(name, value, context)
--- a/lib/brakeman/processors/alias_processor.rb
+++ b/lib/brakeman/processors/alias_processor.rb
@@ -347,6 +347,13 @@ class Brakeman::AliasProcessor < Brakeman::SexpProcessor
           exp = safe_literal(exp.line)
         end
       end
+    when :index_by
+      # Convert %w[a b].index_by(&:itself) to {"a" => "a", "b" => "b"}
+      if array? target and all_literals? target and
+          node_type? first_arg, :block_pass and first_arg[1] == s(:lit, :itself)
+        hash_body = target.each_sexp.flat_map { |e| [e, e.deep_clone] }
+        exp = s(:hash, *hash_body).line(exp.line)
+      end
     end
 
     exp
--- a/test/tests/alias_processor.rb
+++ b/test/tests/alias_processor.rb
@@ -1513,4 +1513,26 @@ class AliasProcessorTests < Minitest::Test
     # Does not raise because the error is caught and handled by the Tracker
     assert Brakeman::AliasProcessor.new(BrakemanTester.new_tracker).process s(:block, s(:attrasgn, s(:call, nil, :x), :"[]!", s(:lit, 1), s(:lit, 2)))
   end
+
+  def test_index_by_itself
+    assert_alias '{"a" => "a", "b" => "b"}', <<-RUBY
+      x = %w[a b].index_by(&:itself)
+      x
+    RUBY
+  end
+
+  def test_index_by_itself_with_freeze
+    assert_alias '{"a" => "a", "b" => "b"}', <<-RUBY
+      x = %w[a b].index_by(&:itself).freeze
+      x
+    RUBY
+  end
+
+  def test_index_by_without_itself
+    # Should not convert if not &:itself
+    assert_alias '%w[a b].index_by(&:to_s)', <<-RUBY
+      x = %w[a b].index_by(&:to_s)
+      x
+    RUBY
+  end
 end
--- a/test/tests/constants.rb
+++ b/test/tests/constants.rb
@@ -278,4 +278,32 @@ class ConstantTests < Minitest::Test
     end
     INPUT
   end
+
+  def test_constants_index_by_itself
+    # index_by(&:itself) on an array of literals should be converted to a hash
+    value = s(:call,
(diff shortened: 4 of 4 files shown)


lib/brakeman/tracker/constants.rb
```diff
@@ -105,6 +105,15 @@ def add name, value, context = nil
         value = value.target
       end
 
+      # Convert %w[...].index_by(&:itself) to a hash literal
+      if call? value and value.method == :index_by and
```
presidentbeef, line 109: Is this necessary? Seems like a very specific case to add to a very general method.

allard: Are you talking about these specific lines or the PR in general? I have this specific issue and we discussed in Issue https://github.com/presidentbeef/brakeman/issues/2018. It would be great if Brakeman wouldn't flag this issue.

presidentbeef: I mean these specific lines in this specific method

allard:
> I mean these specific lines in this specific method

I dug into this and the answer is more nuanced than I originally thought. **You're right that for many cases the `Constants#add` change isn't needed**, but it's needed for cross-file constant usage. Here's what I found.

### Two code paths add a constant to the tracker

1. `BaseProcessor#process_cdecl` — runs first, during initial controller/model/lib parsing. Stores `exp.rhs` **raw** (without alias processing).
2. `AliasProcessor#process_cdecl` — runs later, during data flow analysis. Calls `process exp.rhs` first, so my `process_call` change converts `index_by(&:itself)` to a hash here.

Result: after both passes, the tracker holds **two** entries for the same constant — the first is the raw `:call`, the second is the converted `:hash`. `find_constant` returns the **first** match, so `get_simple_value` rejects it as non-literal and returns `nil`.

### Test fixture results

I built a minimal Rails fixture with `TEMPLATE_IDS = %w[...].index_by(&:itself).freeze` and `File.read("/tmp/templates/#{TEMPLATE_IDS[params[:id]]}.html")`.

**Same-file** (constant defined and used in the same controller):

| Configuration | False positive? |
|---|---|
| No fix | ❌ Yes |
| Alias processor only | ✅ No |
| `Constants#add` only | ❌ Yes |
| Both | ✅ No |

**Cross-file** (constant defined in `lib/template_registry.rb`, used in a controller):

| Configuration | False positive? |
|---|---|
| No fix | ❌ Yes |
| Alias processor only | ❌ Yes |
| `Constants#add` only | ✅ No |
| Both | ✅ No |

### Why the difference

In the same-file case, `AliasProcessor#process_cdecl` also caches the converted value in its local `env` (line 883: `env[match] = get_rhs(exp)`). When the constant is later referenced in the same scope, `replace` (line 86) finds it in `env` *before* falling back to `tracker.constant_lookup`. So the tracker's stored value doesn't matter.

In the cross-file case, the constant is never in the consuming file's `env`, so resolution goes straight to `tracker.constant_lookup` → `get_simple_value`, which requires the stored value to actually be a literal. Without the `Constants#add` change, `find_constant` returns the raw `:call` from `BaseProcessor` → returns `nil` → unresolved → false positive.

### Scope

The case in #2018 is same-file, so as you suspected, only the alias processor change is strictly required for that specific issue. Happy to drop the `Constants#add` change and its tests if you'd prefer to keep the PR minimal — the cross-file improvement is real but it's outside the scope of #2018.

Let me know which way you'd like to go.

presidentbeef:
Yes we should stick to the minimal change here.

Handling the cross-file issue should be fixed more generally, not with a single special case.

PR #2022, opened 2026-05-06
Remove command injection false positives for splatted args

### LINKED ISSUE #2017: False positive: Command injection
### Background
Brakeman version: v8.0.4
Rails version: 8.0.2.1
Ruby version: 3.4.1

Link to Rails application code: sorry, private repo
#### False Positive

*Full* warning from Brakeman: 
```
== Warnings ==

Confidence: Medium
Category: Command Injection
Check: Execute
Message: Possible command injection
Code: Open3.capture3("git", "diff", "origin/main", "origin/#{release_branch}".shellescape, "db/fixtures/masters/festivals.csv")
File: app/services/event_release_service/festivals/extract_release_targets.rb
Line: 12
```

Relevant code:

```ruby
def self.call(release_branch)
  ...
  diff_output, _error, _status = Open3.capture3('git', 'diff', 'origin/main', "origin/#{release_branch}".shellescape, 'db/fixtures/masters/festivals.csv')
```

Brakeman treats dynamic string segments passed into Open3.capture3 as if they might be executed by a shell. Here, Open3.capture3 is called with multiple separate string arguments. In Ruby, that form uses spawn-style execution: the program is invoked with a discrete argv list and no shell is spawned by default. The interpolated value is therefore passed as a single argument to git (the refspec origin/<branch>), not parsed as shell syntax. Characters that would be dangerous in a shell string (spaces, ;, |, `, $(), etc.) do not gain shell metacharacter meaning in this API shape; they are only part of the ref string git receives.

So the finding conflates “string built with interpolation” with “user input concatenated into a shell command line.” The latter is command injection; the former, in multi-argument Open3.capture3, is argument passing to git only. Any abuse is limited to what git diff does with a malformed ref name, not arbitrary command execution via the shell.

JacobEvelyn:
This commit expands `check_execute.rb` to no longer warn about splatted args being passed to a variety of additional methods (such as `capture3` and `popen3`), mirroring how the check already treated `system` and `exec`.

This commit also expands that check to correctly handle when environment variables are passed to some of these methods via a hash as the first argument.

Fixes #2017

---

Note: #2017 was already marked as fixed, but the actual issue reported by @marocchino was not actually addressed in [the fix](https://github.com/presidentbeef/brakeman/commit/169a7d4a800ca13188e74e871fb8ade3ab349470):

> Brakeman treats dynamic string segments passed into Open3.capture3 as if they might be executed by a shell. Here, Open3.capture3 is called with multiple separate string arguments. In Ruby, that form uses spawn-style execution: the program is invoked with a discrete argv list and no shell is spawned by default. The interpolated value is therefore passed as a single argument to git (the refspec origin/), not parsed as shell syntax. Characters that would be dangerous in a shell string (spaces, ;, |, `, $(), etc.) do not gain shell metacharacter meaning in this API shape; they are only part of the ref string git receives.

> So the finding conflates “string built with interpolation” with “user input concatenated into a shell command line.” The latter is command injection; the former, in multi-argument Open3.capture3, is argument passing to git only. Any abuse is limited to what git diff does with a malformed ref name, not arbitrary command execution via the shell.

### DIFF
--- a/lib/brakeman/checks/check_execute.rb
+++ b/lib/brakeman/checks/check_execute.rb
@@ -55,6 +55,13 @@ class Brakeman::CheckExecute < Brakeman::BaseCheck
     first_arg = call.first_arg
     failure = nil
 
+    # All spawn-family methods optionally accept an env hash as the first
+    # argument. Strip it so first_arg always points to the command name.
+    if hash?(first_arg)
+      args.delete_at(1)
+      first_arg = args[1]
+    end
+
     case call.method
     when :popen
       # Normally, if we're in a `popen` call, we only are worried about shell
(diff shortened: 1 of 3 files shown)


JacobEvelyn:
@presidentbeef just nudging this in case you missed it

JacobEvelyn:
@presidentbeef pinging you here again

presidentbeef:
Fix looks reasonable but is it really for splatted args? None of the tests use `*args`.

JacobEvelyn:
Ah, I guess splatted wasn't the right term. I meant when you pass in multiple args to form a shell command, e.g. `system("echo", "hello")` instead of `system("echo hello")`. Some codebases achieve that via splatting (`command = ["echo", "hello"]; system(*command)`) but you're right that the splatting is irrelevant here.

Issue #1972, opened 2025-10-23
`Brakeman::SexpProcessor#process`: Type should be a Symbol


Jack12816:
Hey there 👋 first things first: thank you for maintaining this great tool/gem! ❤️ 

### Background

Brakeman version: 7.1.0 (works with 7.0.2)
Rails version: 8.0.3
Ruby version: ruby 3.4.7 (2025-10-08 revision 7a5688e2a2) +PRISM [x86_64-linux]

Issue occurs with `--prism` AND `--no-prism` - seems unrelated.

### Issue

```
/usr/local/bundle/gems/brakeman-7.1.0/lib/ruby_parser/bm_sexp_processor.rb:70
'Brakeman::SexpProcessor#process': Type should be a Symbol
```

<details>
  <summary>Full stack trace</summary>

```
[Notice] Using configuration in /app/config/brakeman.yml
[Notice] Using Prism parser
Loading scanner...
Processing application in /app
Processing gems...
[Notice] Using Prism parser
Parsing Gemfile
[Notice] Detected Rails 8 application
(Processing gems) Duration: 0.007603007 seconds
Processing configuration...
[Notice] Using Prism parser
Parsing config/environment.rb
[Notice] Using Prism parser
Parsing config/application.rb
[Notice] Using Prism parser
Parsing config/environments/production.rb
[Notice] Escaping HTML by default
(Processing configuration) Duration: 0.002594072 seconds
Finding files...
(Finding files) Duration: 0.382053285 seconds
Parsing files...
[Notice] Using Prism parser
Parsing [...]
(Parsing files) Duration: 0.663424268 seconds
Detecting file types...
(Detecting file types) Duration: 0.007480011 seconds
Processing initializers...
(Processing initializers) Duration: 0.002072824 seconds
Processing libs...
(Processing libs) Duration: 0.014954371 seconds
Processing routes...
[Notice] Using Prism parser
Parsing config/routes.rb
(Processing routes) Duration: 0.002886525 seconds
Processing templates...
(Processing templates) Duration: 0.027191228 seconds
Processing data flow in templates...
[.. unrelated ..]
Rendering management/users/logins (["Management::UsersController#logins"])
Error when processing management/users/logins: Error when processing management/users/logins: Error when processing management/users/logins: Error when processing management/users/logins: Error when processing management/users/logins: Error when processing management/users/logins: Error when processing management/users/logins: Error when processing management/users/logins: undefined method 'empty?' for nil
[...]

PR #2023, opened 2026-05-26
Fix HAML crash on .raw calls

FFederi:
I tried fixing https://github.com/presidentbeef/brakeman/issues/1972.
Seems like calls like `login.user_agent.raw` were being wrongly identified as calls to Rails' `raw()` view helper.
Let me know what you think!


presidentbeef:
Wow, thanks for fixing this. I thought it was something way more complicated.

PR #2025, opened 2026-06-03
Fix SQL injection false positive for compact_blank/compact on permitted params

### LINKED ISSUE #1885: False positive when methods chained on permitted params
### Background

Brakeman version: 6.2.2
Rails version: 8.0.0
Ruby version: 3.3.5

#### False Positive

*Full* warning from Brakeman:
```
Confidence: High
Category: SQL Injection
Check: SQL
Code: MyModel.where(params.permit(:foo, :bar).slice(:foo, :bar).compact_blank)
```

Relevant code:

```ruby
MyModel.where(params.permit(:foo, :bar).slice(:foo, :bar)) # no warning
MyModel.where(params.permit(:foo, :bar).slice(:foo, :bar).compact_blank) # warning
```

The `compact_blank` method returns a new `ActionController::Parameters` object, but that new object retains the `permitted` status from our earlier call to `permit`.  There is a list of permitted method calls defined [here](https://github.com/presidentbeef/brakeman/blob/88994b8fe026f38c46e15988559ec6574d37895a/lib/brakeman/checks/check_sql.rb#L308) but `compact_blank` is not listed there.  It probably shouldn't be, because `compact_blank` on its own does not make the call safe, but being chained after `permit` ought to.  

Is there a way to mark methods that are chained after `permit` as safe?

arpitjain099:
Fixes #1885.

### The false positive

```ruby
MyModel.where(params.permit(:foo, :bar).slice(:foo, :bar))               # no warning
MyModel.where(params.permit(:foo, :bar).slice(:foo, :bar).compact_blank) # false SQL injection warning
```

`check_sql.rb` keeps a list of methods that, when chained off `params`, preserve `permit`'s permitted/safe status: `:permit, :slice, :to_h, :to_hash, :symbolize_keys`. The check only looks at the outermost method in the chain, so when the chain ends in `compact_blank` (or `compact`), the call is treated as unsafe even though everything before it was already permitted.

`compact_blank` and `compact` return a new `ActionController::Parameters` that retains the permitted status of the receiver and only removes blank/nil values. They do not reintroduce user-controlled keys or otherwise re-taint the parameters, so a chain that was already safe stays safe.

### The fix

Add `:compact` and `:compact_blank` to that allowlist in `lib/brakeman/checks/check_sql.rb`. This mirrors how `:slice`, `:to_h`, and `:symbolize_keys` are already handled (those also only suppress the warning based on the outermost method).

A genuinely unsafe query (a raw interpolated string) still warns.

### Tests

Following the convention in CONTRIBUTING.md, I modified the `rails8` test app to reproduce the issue and added tests in `test/tests/rails8.rb`:

- `test_sql_injection_compact_blank_false_positive` and `test_sql_injection_compact_false_positive` - `assert_no_warning` for the two chained-method cases.
- `test_sql_injection_still_warns_on_interpolation` - `assert_warning` confirming an interpolated `where(...)` is still reported.
- Bumped the expected generic-warning count from 9 to 10 for the one intentional new (real) warning.

Reverting the one-line check change makes the two false-positive tests (and the warning-count test) fail, so the tests are meaningful rather than vacuous. `ruby test/tests/rails8.rb` is green, and `rails6`/`rails52` still pass with no regressions.

### DIFF
--- a/lib/brakeman/checks/check_sql.rb
+++ b/lib/brakeman/checks/check_sql.rb
@@ -311,7 +311,7 @@ class Brakeman::CheckSQL < Brakeman::BaseCheck
     end
 
     if request_value? arg
-      unless call? arg and params? arg.target and [:permit, :slice, :to_h, :to_hash, :symbolize_keys].include? arg.method
+      unless call? arg and params? arg.target and [:permit, :slice, :to_h, :to_hash, :symbolize_keys, :compact, :compact_blank].include? arg.method
         # Model.where(params[:where])
         arg
       end
--- a/test/apps/rails8/app/controllers/users_controller.rb
+++ b/test/apps/rails8/app/controllers/users_controller.rb
@@ -83,6 +83,18 @@ class UsersController < ApplicationController
     @users = User.all
   end
 
+  def permitted_compact_blank
+    # compact_blank/compact return permitted Parameters, so chaining them
+    # after permit/slice should not raise a SQL injection warning.
+    User.where(params.permit(:foo, :bar).slice(:foo, :bar).compact_blank)
+    User.where(params.permit(:foo, :bar).slice(:foo, :bar).compact)
+  end
+
+  def still_unsafe_sql
+    # A raw interpolated string is still unsafe and should warn.
+    User.where("name = '#{params[:name]}'")
+  end
+
   private
     # Use callbacks to share common setup or constraints between actions.
     def set_user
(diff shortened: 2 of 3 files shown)


presidentbeef:
Please fix the merge conflicts and then I'll merge this.

arpitjain099:
Rebased onto main and resolved the conflict in the rails8 test controller (kept your `chained_raw` plus the two new test methods); the fix itself is unchanged. I could not run the suite locally (Ruby 2.6 here, brakeman needs 3.2+), so CI will confirm. Ready when you are, thanks!

presidentbeef:
Tests failed because the code moved but the line numbers in the tests were not updated.

arpitjain099:
Fixed - after the merged-in method shifted the fixture, the still-warns case moved to line 95; updated the assertion (fingerprint unchanged). And I got a 3.2+ Ruby set up locally now per your nudge, so the rails8 suite runs green here (23 runs, 0 failures). Thanks!

PR #2027, opened 2026-06-04
Fix RenderInline false positive for partial local named text

### LINKED ISSUE #1915: False positive warning when rendering a partial that has a local called `text`
### Background

Brakeman version: 7.0.0
Rails version: 7.1.5.1
Ruby version: 3.3.7

### Issue

The following code:

```erb
<%= render(
  "inventory_unit/shared/inventory_record",
  text: params.dig(:inventory_unit, :inventory_record, :comment)
) %>
```

Produces:

```
Confidence: High
Category: Cross-Site Scripting
Check: RenderInline
Message: Unescaped parameter value rendered inline
Code: render(text => params.dig(:inventory_unit, :inventory_record, :comment), {})
File: app/inventory/views/inventory_unit/new.html.erb
Line: 86
```

However, changing the parameter name from `text` to `comment` doesn't produce a warning:

```erb
<%= render(
  "inventory_unit/shared/inventory_record",
  comment: params.dig(:inventory_unit, :inventory_record, :comment)
) %>
```

arpitjain099:
Fixes #1915

## Problem

Rendering a partial with a local whose name happens to match a Rails render type (most commonly `text`) produces a false positive `RenderInline` / Cross-Site Scripting warning:

```erb
<%= render("shared/inventory_record", text: params[:comment]) %>
```

reports:

```
Confidence: High
Category: Cross-Site Scripting
Check: RenderInline
Message: Unescaped parameter value rendered inline
```

Renaming the local to anything that is not a render type name (for example `comment:`) makes the warning disappear, which confirms it is a parsing issue rather than a real finding.

## Root cause

`Brakeman::BaseProcessor#find_render_type` first sets the render type from the leading positional argument (a string/symbol partial or action name). It then iterates the options hash and, for any key whose name is in the render-type set (`:text`, `:inline`, `:action`, ...), overrides the type and value with that key.

When the type was already established by a positional partial name, a hash key like `text:` is just a local passed to the partial, not a `render text:` directive. The override reclassified the partial render as a `:text` render and pointed its value at the user input, so `RenderInline` flagged it.

## Fix

Track whether the render type was determined by a leading positional argument. Only treat a matching options-hash key as a render-type directive when the type was not already set positionally. In the partial case the key is now retained as a local, matching Rails behavior.

## Verification

Before (on `main`), against a minimal app:

```
RenderInline | Cross-Site Scripting | line 1 | Unescaped parameter value rendered inline
TOTAL: 1
```

After:

```
TOTAL: 0
```

True positives are unaffected. `render(text: params[:x])` and `render(inline: params[:y])` (no positional argument) still report inline XSS.

A regression test was added in `test/tests/rails5.rb` (`test_render_inline_partial_local_named_text`) with a fixture line in `test/apps/rails5/app/views/widget/show.html.erb`. The test uses `assert_no_warning` and fails on the unpatched parser, passes with the fix. The full suite count is unchanged aside from the one added test (pre-existing environmental failures on Ruby 4.0 are present on `main` as well).

### DIFF
--- a/lib/brakeman/processors/base_processor.rb
+++ b/lib/brakeman/processors/base_processor.rb
@@ -227,6 +227,11 @@ class Brakeman::BaseProcessor < Brakeman::SexpProcessor
     end
 
     #Look for render :action, ... or render "action", ...
+    #When the type is determined by a leading positional argument, a key in the
+    #options hash that happens to match a render type name (e.g. `text:`) is a
+    #local passed to the partial/action, not a render type directive.
+    type_from_positional = false
+
     if string? first_arg or symbol? first_arg
       if @current_template and @tracker.options[:rails3]
         type = :partial
@@ -235,15 +240,18 @@ class Brakeman::BaseProcessor < Brakeman::SexpProcessor
         type = :action
         value = first_arg
       end
+      type_from_positional = true
     elsif first_arg.is_a? Symbol or first_arg.is_a? String
       type = :action
       value = Sexp.new(:lit, first_arg.to_sym).line(call.line)
+      type_from_positional = true
     elsif first_arg.nil?
       type = :default
     elsif not hash? first_arg
       # Maybe do partial if in view?
       type = :action
       value = first_arg
+      type_from_positional = true
     end
 
     types_in_hash = Set[:action, :file, :inline, :js, :json, :nothing, :partial, :template, :text, :update, :xml]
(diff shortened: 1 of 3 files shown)


presidentbeef:
If `render` has a first argument, even if it's `nil`, then I think the render type would be determined by that argument.

So if `type` is set, we should never override it from the hash argument. I think. So a better fix might just be to skip pulling it out of the hash argument entirely if `type` is already set. That also avoids building up `rest` manually (if type is set) because `rest` would be `last_arg` if it's a hash.

arpitjain099:
Switched to your suggested approach: when the type is already set by a positional argument, the options hash is treated as locals (`rest = last_arg`) rather than scanning it for a render type, so the extra flag is gone and `rest` is no longer rebuilt by hand. The regression test (render of a partial with a `text:` local) covers the original false positive. I couldn't run the suite locally (Ruby 2.6 here, brakeman needs 3.2+), so CI will confirm. Thanks for the steer!

presidentbeef:
You should probably figure out how to upgrade Ruby wherever you are testing this 😉

arpitjain099:
Got a 3.2+ Ruby set up locally per your nudge 😄 - the rails5 suite passes green (70 runs, 311 assertions, 0 failures) with the revised approach. Thanks again!

PR #2028, opened 2026-06-10
Add quote_schema_name to safe quote method list

kiskoza:
The quote_schema_name method was introduced in Rails 5.0 and since then it almost always just calls quote_column_name which is already in the safe list.

https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/PostgreSQL/Quoting.html#method-i-quote_schema_name

I added the new test to the rails4 folder as I found similar tests there, but happy to move them into a better place if I get some guidance.


presidentbeef:
Thanks!

PR #2032, opened 2026-07-10
Fix glob pattern to include specific extension

okuramasafumi:
This could make a significant performance improvement when directories with too many files (`.node_modules`, for example).


presidentbeef:
Directly-related tests are failing.

Also, this is in a non-default branch of the code, so gains will be only be for users of `--follow-symlinks`.

okuramasafumi:
@presidentbeef After checking our configuration, we don't use `--follow-symlinks` option. Now I don't know why it was so slow, but adding `skip_files` to `brakeman.yml` solved it. I'm closing this PR, but I'd like to revisit it if I found the root cause.

PR #2034, opened 2026-08-12
Fix EOL date for Rails 8.0

yeaseul-kim:
## Background

`RAILS_EOL_DATES` lists Rails 8.0.x as going end-of-life on `2026-10-07`, but the correct date is **2026-11-07**. Rails 8.0.0 was released on 2024-11-07 and gets 2 years of security fixes, and both sources already linked above the table agree:

* <https://rubyonrails.org/maintenance> — "8.0.x - Supported until November 7, 2026"
* <https://endoflife.date/rails> — Rails 8.0 security support ends 07 Nov 2026

## Impact

`EOLCheck` warns 60 days ahead of the EOL date, so the wrong month makes Brakeman report Rails 8.0 as approaching end of life a month early. Apps on Rails 8.0 have been getting this since 2026-08-08:

```
Confidence: Weak
Category: Unmaintained Dependency
Check: EOLRails
Message: Support for Rails 8.0.5.1 ends on 2026-10-07
File: Gemfile.lock
```

With the corrected date it starts on 2026-09-08 instead.

## Change

```diff
-    ['8.0.0', '8.0.99'] => Date.new(2026, 10, 7),
+    ['8.0.0', '8.0.99'] => Date.new(2026, 11, 7),
```

The neighbouring entries check out against the same two sources, so they are left alone.

No test is added, matching how the other EOL dates were introduced — this is a data correction rather than a behavioural change. The existing `EOLRails` assertions all cover Rails versions that are already past EOL, and giving Rails 8.0 the same treatment would mean moving the Rails 8 scan's `Date.today` stub past 2026-11-07, which would churn that application's expected warning counts.


presidentbeef:
Huh. Does look like it was just a mistake on the original entry.

PR #2039, opened 2026-08-17
Skip top-level vendor directory before recursive globbing

ronocod:
👋 

Been digging into some performance improvements based on our usage at [Tines](https://www.tines.com/). Got a couple of PRs coming over the next while, starting with this one which is the simplest.

Our `vendor` directory is quite big, and `skip_vendor` says it should be skipped by default but in practice it's not - the `relative_path.include?('vendor/')` only covers paths under the directory, but not the `vendor` directory itself. Excluding the directory here saves a lot of work dealing with it later in the filtering pipeline.

Benchmarked with `hyperfine --warmup 2 --runs 5`:

```
# this branch
Benchmark 1: bundle exec ./bin/brakeman --path ../tines/backend --gemfile Gemfile --no-pager --no-progress --no-exit-on-warn --no-exit-on-error
  Time (mean ± σ):     17.438 s ±  0.211 s    [User: 17.030 s, System: 3.996 s]
  Range (min … max):   17.207 s … 17.716 s    5 runs

# main
Benchmark 1: bundle exec ./bin/brakeman --path ../tines/backend --gemfile Gemfile --no-pager --no-progress --no-exit-on-warn --no-exit-on-error
  Time (mean ± σ):     20.805 s ±  0.104 s    [User: 17.948 s, System: 5.784 s]
  Range (min … max):   20.679 s … 20.957 s    5 runs
```


lib/brakeman/app_tree.rb
```diff
@@ -260,11 +263,17 @@ def reject_global_excludes(paths)
     end
 
     def in_engine_paths?(path)
-      @engine_paths.any? { |p| path.absolute.include?(p) }
+      @engine_paths.any? do |p|
+        path.absolute.include?(p) ||
+          expand_path(p).start_with?("#{path.absolute}#{File::SEPARATOR}")
+      end
```
ronocod, line 269: Now that `path.absolute` can just be `/app/vendor` rather  than a file path, we need to ensure that supplied engine/lib paths like `vendor/my_engine` still match it.

presidentbeef: I'm not 100% following the reasoning of why this is needed, but I'm wary of the `expand_path...` branch because I'm pretty sure that's going to be the hot path.

Can you explain a little bit more?

ronocod: Fair point about the hot path, I've moved this expansion out of there now so it's done once and cached.

The reason this logic is here is for cases like:
```
skip_vendor: true
engine_paths: ["vendor/bundle/gems/my_engine"]
```

Before this change, `reject_global_excludes` inside `top_directories_pattern` always returned `false` for the path `/repo/vendor` so it always got passed through to `Dir.glob` at the end of `glob_files`, which would subsequently iterate through all files under that directory. Later, in `select_files`, we run `reject_global_excludes` and then reject them all anyway. This happened even if `skip_vendor` was enabled.

Now, we reject the top-level `/repo/vendor` path inside `reject_global_excludes` in `top_directories_pattern`, which means we never `Dir.glob` to enumerate all those files, and they subsequently don't get passed through to `select_files`. The problem is that we shouldn't reject `/repo/vendor` early if an engine path like `vendor/bundle/gems/my_engine` is configured as it lives under there, meaning we need to do the `Dir.glob` expansion to figure out which files under `/repo/vendor` to keep and which to exclude. To make that happen, `in_engine_paths?("/repo/vendor")` needs to return `true`, which means we can't just check if it includes `"vendor/bundle/gems/my_engine"` (the existing check), we also have to check if the expanded `"/repo/vendor/bundle/gems/my_engine"` starts with `path` too.

Hope that makes sense, happy to dig into this more if you want!

PR #2044, opened 2026-08-31
Fix frozen string error

shaicoleman:
When frozen strings are enabled via `RUBYOPT='--enable=frozen-string-literal'` brakeman gives the following error:
```
gems/brakeman-8.0.6/lib/brakeman/warning.rb:178:in 'String#gsub!': can't modify frozen String: "[Format Error]" (FrozenError)
  from gems/brakeman-8.0.6/lib/brakeman/warning.rb:178:in 'Brakeman::Warning#format_with_user_input'
  from gems/brakeman-8.0.6/lib/brakeman/report/report_text.rb:205:in 'Brakeman::Report::Text#format_code'
  from gems/brakeman-8.0.6/lib/brakeman/report/report_text.rb:177:in 'Brakeman::Report::Text#format_line'
```


presidentbeef:
Fix is fine, but I think the real issue is whatever is causing the format error. Feel free to open a separate issue if you can narrow down the problematic code.

PR #2050, opened 2026-09-18
Fix CheckValidationRegex overly broad ignore behaviour

eliotsykes:
This fixes an issue where ignoring one `CheckValidationRegex` warning would ignore all `CheckValidationRegex` warnings for the same file, as they all shared the same fingerprint.

For example, take this model with 2 format validations that generates two `CheckValidationRegex` warnings:

```ruby
class Person < ApplicationRecord
  validates :first_name, :format => /\w+/
  validates :email, :format => /@/
end
```

Prior to this change, if you ignored one warning, the other warning would also be unintentionally ignored as both warnings shared the same fingerprint.

After this change, ignoring one of these warnings will only ignore that warning. It makes `CheckValidationRegex` warning fingerprints more unique by taking into account the vulnerable `Sexp` `code`.


lib/brakeman/checks/check_validation_regex.rb
```diff
@@ -92,7 +92,8 @@ def check_regex value, validator
       :message => msg("Insufficient validation for ", msg_code(get_name validator), " using ", msg_code(regex.inspect), ". Use ", msg_code("\\A"), " and ", msg_code("\\z"), " as anchors"),
       :line => value.line,
       :confidence => :high,
-      :cwe_id => [777]
+      :cwe_id => [777],
+      :code => validator
```
eliotsykes, line 96: This change is the essence of this PR. Setting `:code` on the warning makes the fingerprint adequately unique.

test/tests/sarif_output.rb
```diff
@@ -121,12 +121,8 @@ def test_results_shape
...
   end
 
-  def test_with_ignore_has_one_suppressed_finding
-    assert_equal(
-      1,
-      @@sarif_with_ignore.dig('runs', 0, 'results').
-        select { |f| f['suppressions'] }.count
-    )
+  def test_with_ignore_has_suppressed_findings
+    refute_empty @@sarif_with_ignore.dig('runs', 0, 'results').select { |f| f['suppressions'] }
```
eliotsykes, line 125: Adding another entry to `test/apps/rails4/config/brakeman.ignore` incremented this from 1 to 2 suppressions. I've modified the test to make it more resilient to future additions to `brakeman.ignore` whilst trying to preserve its intent.

eliotsykes:
`test/tests/rails2.rb` is failing under Ruby 4.0.7 on CI on a couple of recent PRs. One of those PRs is a one character change which makes me wonder if this is usual. Is this expected?

presidentbeef:
Tests are fixed on main by #2048

eliotsykes:
Thanks @presidentbeef, do you have a preferred way to rerun the failed tests on a PR? I was thinking of rebasing and force-pushing to (hopefully) trigger a re-run.

presidentbeef:
Rebasing and pushing would be the preferred approach

